PR #35, opened 2015-01-29
Make the README conform with the name used on the package

diasdavid:


fantix:
Sorry for being late - got some problem with email notification with GitHub. Thanks for the fix!

PR #37, opened 2015-01-31
Update with language changes

axelmagn:
These are just a few quick refactors to keep pace with changes in the rust language.

 * Updated attribute `deriving` -> `derive`
 * Refactored `std::io` -> `std::old_io`

I was unable to solve some failures surrounding the `SocketAddr::from_str` function.  For some reason it is not recognized.  So this doesn't bring `zmq.rs` to build-ready, but hopefully it takes care of some of the busy-work.


fantix:
Thanks a lot! I'll take it and see how it helps :)

PR #54, opened 2020-06-28
Implemented round robin and other spec related features for ReqSocket

Alexei-Kornienko:
Now this socket seems to be spec compliant. Need to test it properly and benchmark overall code

### DIFF
--- a/src/req_rep.rs
+++ b/src/req_rep.rs
@@ -113,6 +113,11 @@ impl SocketFrontend for ReqSocket {
     }
 
     async fn bind(&mut self, endpoint: &str) -> ZmqResult<()> {
+        if self._accept_close_handle.is_some() {
+            return Err(ZmqError::Other(
+                "Socket server already started. Currently only one server is supported",
+            ));
+        }
         let stop_handle = util::start_accepting_connections(endpoint, self.backend.clone()).await?;
         self._accept_close_handle = Some(stop_handle);
         Ok(())
--- a/src/req_rep.rs
+++ b/src/req_rep.rs
@@ -14,7 +14,7 @@ use std::time::Duration;
 use tokio::stream::StreamExt;
 
 struct ReqSocketBackend {
-    pub(crate) peers: Arc<DashMap<PeerIdentity, Peer>>,
+    pub(crate) peers: DashMap<PeerIdentity, Peer>,
     pub(crate) round_robin: SegQueue<PeerIdentity>,
     pub(crate) current_request_peer_id: AtomicCell<usize>,
 }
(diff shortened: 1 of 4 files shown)


src/req_rep.rs
```diff
@@ -119,6 +151,7 @@ impl MultiPeer for ReqSocketBackend {
                 _io_close_handle: stop_handle,
```
malor, line 151: Hmm, why do we need to synchronise access to `in_queue_receiver`? Isn't the receiving end of this channel only used in `ReqSocket::recv()`?

src/req_rep.rs
```diff
@@ -119,6 +151,7 @@ impl MultiPeer for ReqSocketBackend {
...
             },
```
malor, line 152: Don't we have at most 1 outstanding request in ZMQ_REQ sockets? Is there a reason to have channels with size > 1?

src/req_rep.rs
```diff
@@ -25,21 +28,43 @@ pub struct ReqSocket {
...
+                    let frames = vec![
+                        "".into(), // delimiter frame
+                        message,
+                    ];
+                    peer.send_queue
+                        .send(Message::MultipartMessage(frames))
+                        .await?;
+                    self.backend
+                        .current_request_peer_id
+                        .store(self.backend.peers.hash_usize(&next_peer_id));
```
malor, line 61: This should be fine, but `hash_usize()` is *probably* not on par with cryptographic hash functions like SHA-1, so I'd expect the probability of collisions to be higher. IIUC, the only reason you are going for using of the hash function in the first place is that you want to use an atomic, but peer identity is a vector of u8's and can't be used as one. Perhaps, an alternative would be to map peer identity to an incrementing integer and use that instead.

src/req_rep.rs
```diff
@@ -78,14 +103,21 @@ impl SocketFrontend for ReqSocket {
...
+                peers: DashMap::new(),
+                round_robin: SegQueue::new(),
+                current_request_peer_id: AtomicCell::new(0),
             }),
             _accept_close_handle: None,
             current_request: None,
         }
     }
 
     async fn bind(&mut self, endpoint: &str) -> ZmqResult<()> {
```
malor, line 115: Hmm, shouldn't a ZMQ_REQ socket only have `connect()`, and not `bind()`?

Alexei-Kornienko: No all zmq sockets can work both as client or server. It depends on actual use case which variant is used

Alexei-Kornienko:
Agreed on most things on a call. Possibly will return to some remaining issues later in another request

PR #59, opened 2020-09-17
Added github-actions CI and removed travis

TheButlah:
This switches from travis CI to github actions.

Added:
- main-ci.yml, which describes the main CI workflow of the project. Deployment CI will likely go in a separate workflow.
- Four CI jobs: check that code compiles, check that tests pass, check that code is formatted, check for lints. The lint job can fail and will produce warnings instead of blocking a merge.
- Added a rustfmt.toml file. While technically unnecessary, it announces that the project is formatted with rustfmt and any deviations from the default config.

Removed:
- travis config yaml

The following in the repo settings will require the status checks to pass before merging to master:

1. Go to settings, branches, then add rule.
2. Enter the name of the branch as `master`
3. Select "Require status checks to pass before merging", and check the boxes for the 4 newly added status checks (note that even though the lint job is marked as required, its not actually going to block a merge due to the settings in the config.yml file.)
4. Check "include administrators", to avoid accidentally merging stuff to master that fails the checks (this one is less important than steps 1-3)

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -56,7 +56,7 @@ jobs:
           command: fmt
           args: --all -- --check
 
-  clippy:
+  lint:
     name: Linting
     continue-on-error: true
     runs-on: ubuntu-latest
@@ -65,7 +65,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: nightly
           override: true
           components: clippy
       - name: Cache cargo dependencies
@@ -73,7 +73,7 @@ jobs:
         with:
           path: ~/.cargo/
           key: cache-cargo-deps-${{ runner.os }}-${{ hashFiles('**/Cargo.lock', '**/Cargo.toml') }}
-      - uses: actions-rs/cargo@v1
+      - uses: actions-rs/clippy-check@v1
         with:
-          command: clippy
-          args: -- -D warnings
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-features
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -1,0 +1,79 @@
+name: Main CI
+on: push
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


Alexei-Kornienko:
I'll take a closer look on this ASAP (most likely on the weekend). Unfortunately I don't have much time to work on this project

Alexei-Kornienko:
@TheButlah Hi, basically I don't care what CI to use as long as it works. Could you please explain in a little more detail what's the profit in switching to github actions?

TheButlah:
To some extent it is personal preference, and the fact that I wanted to add a more comprehensive CI setup (the PR adds a few extra jobs that the current travis setup doesn't do, and because github actions or gitlab are my CIs of choice, I did it in GH actions). But here are some less arbitrary reasons:

1. Github actions integrates better with github out of the box. This is because there is a dedicated "actions" tab at the top next to the "pull requests" tab that displays all of the CI without going to a different service like travis.
2. The status checks show up as a bit more granular, since each job in the github workflow is a separate status check, rather than a single status check being "travis". So for example if I'm a contributor and my code compiles but my tests fail, I see a green check box next to the compile job, and a red x next to the tests job. Likewise, the "checks" tab in the PR is more descriptive.
3. A common alternative is to set up github bots that manage things for you and report when builds fail to users. There are many good bot setups that other repos use, lang/rust for example makes extensive use of bots, but its a lot more complicated than the out of the box tools you get with github actions, and not clear to me if it adds anything more.
4. There is a really good ecosystem already for rust and github actions, which I suspect is a stronger one than what exists for travis. See https://github.com/actions-rs/. It has code coverage, linting, formatting, minimum stable rust version checks, and more.

Also, it is better for deployment. Consider [this terrain generation project](https://github.com/TheButlah/rust-terrain-gen) I wrote that uses github actions to automatically compile the rust code into web assembly and deploy the code. I can deploy it to github pages to statically host the wasm app, and it happens automatically. If you want to see exactly how that works, see [the deploy.yml file](https://github.com/TheButlah/rust-terrain-gen/blob/master/.github/workflows/deploy.yml)

TheButlah:
**Here's what a PR might look like under the new setup:**

From the PR itself:
image

From the "checks" tab in the PR:
image

Note the built-in integration in the diff view, showing things that caused errors:
image

PR #63, opened 2020-09-22
Fix flaky test and test multiple addresses

TheButlah:
Currently rebased on #62, see that one first

I refactored the pub_sub test to allow for multiple addresses to be tested, and am now testing multiple ipv4 and ipv6 addresses.

I also took the opportunity to make the test less flaky - there is no guarantee that the current time of receipt matches when it was sent, and this caused the test to fail non-deterministically. I fixed it by cloning the payload so that the received value could be directly compared against the expected value in the subscribers.

### DIFF
--- a/tests/pub_sub.rs
+++ b/tests/pub_sub.rs
@@ -6,52 +6,56 @@ use zeromq::prelude::*;
 
 #[tokio::test]
 async fn test_pub_sub_sockets() {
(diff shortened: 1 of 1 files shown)


Alexei-Kornienko:
@TheButlah  Could you please rebase this one?

TheButlah:
@Alexei-Kornienko done

PR #68, opened 2020-09-28
Added Endpoint enum to better encode endpoints for different transport types

TheButlah:
I've added an `Endpoint` enum, along with a `Host` and `Transport` enum. These enums should help us add support for new transport types other than tcp or udp.

Without the enums, I found myself having to check the syntax of the string based endpoint a lot. For example, if its tcp the string should be host:port, but if its ipc it should just be a string address. By consolidating all of these checks into the parse() method of `Endpoint`, we can avoid scattering these parsing and checking steps everywhere in the code, and we can also let people avoid that overhead by directly constructing the Endpoint.

```rust
let str_endpoint: Endpoint = "tcp://example.com:4567".parse()
let enum_endpoint = Endpoint::Tcp("example.com".to_string(), 4567)
```

Things to discuss:
1. I have not figured out the proper types to encode the errors in the `endpoint` module yet - most of them are labeled as `ZmqError::Other`. What do you recommend the error type be?
2. ~~I have made a custom trait called `TryIntoEndpoint` which basically does the same thing as `TryInto<Endpoint, Error=ZmqError>`. I have the trait because I wanted something similar to a trait alias, but that isn't a thing in rust yet (see https://github.com/rust-lang/rust/issues/41517). Is this the right approach?~~
3. ~~I currently accept arguments as `a: impl TryIntoEndpoint` for the user-facing functions accepting an endpoint. This form of static dispatch will get monomorphized in most cases and lead to code bloat, but lets the user pass either strings or `Endpoint` types to these functions, which is a little more ergonomic. However, its really not that hard to pass the argument as `"tcp://example.com:4567".parse()`. It might be worth removing the `impl TryIntoEndpoint` and just doing `a: Endpoint` instead. That might also let us delete the `TryIntoEndpoint` trait entirely, and will lead to faster compile times and smaller binary sizes. Especially if we intend to support WASM as a platform, where binary size really matters, this will be important. What are your thoughts?~~
4. The different variants of the `Endpoint` enum are currently just tuples. For example, I have `Endpoint::Tcp(Host, Port)`. This means that if there are functions that operate exclusively on a particular type of endpoint (lets say `TcpEndpoint`), we currently have to do a match to make sure that we have the right transport type for any given `Endpoint`. It might be worth instead making each variant of the `Endpoint` have a type. For example, we could have both `Endpoint` and `TcpEndpoint`, where `TcpEndpoint` is a tuple struct of `(Host, Port)`, and we would have `Endpoint::Tcp(TcpEndpoint)`. We could also `impl From<TcpEndpoint> for Endpoint` to allow to easily convert from a TcpEndpoint to an Endpoint. The only downside is that it would introduce a lot of additional types, but I think that its worth it for the increase in type safety

### DIFF
--- a/examples/message_broker.rs
+++ b/examples/message_broker.rs
@@ -7,11 +7,11 @@ use zeromq::{Socket, ZmqError};
 async fn main() -> Result<(), Box<dyn Error>> {
     let mut frontend = zeromq::RouterSocket::new();
     frontend
-        .bind("127.0.0.1:5559")
+        .bind("tcp://127.0.0.1:5559")
         .await
         .expect("Failed to bind");
 
-    // let mut backend = zmq_rs::DealerSocket::bind("127.0.0.1:5560")
+    // let mut backend = zmq_rs::DealerSocket::bind("tcp://127.0.0.1:5560")
     //     .await
     //     .expect("Failed to bind");
     //
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,9 @@ enum-primitive-derive = "^0.1"
 dashmap = "^3.11"
 crossbeam = "^0.7"
 uuid = { version = "^0.8", features = ["v4"] }
+regex = "1"
+lazy_static = "1"
+
 
 [dev-dependencies]
 chrono = "^0.4"
--- a/examples/message_client.rs
+++ b/examples/message_client.rs
@@ -7,7 +7,7 @@ use zeromq::{BlockingRecv, BlockingSend};
 async fn main() -> Result<(), Box<dyn Error>> {
     let mut socket = zeromq::ReqSocket::new();
     socket
-        .connect("127.0.0.1:5559")
+        .connect("tcp://127.0.0.1:5559")
         .await
         .expect("Failed to connect");
 
--- a/examples/socket_client.rs
+++ b/examples/socket_client.rs
@@ -7,7 +7,7 @@ use zeromq::{BlockingRecv, BlockingSend};
 async fn main() -> Result<(), Box<dyn Error>> {
     let mut socket = zeromq::ReqSocket::new();
     socket
-        .connect("127.0.0.1:5555")
+        .connect("tcp://127.0.0.1:5555")
         .await
         .expect("Failed to connect");
     println!("Connected to server");
--- a/examples/socket_server.rs
+++ b/examples/socket_server.rs
@@ -5,7 +5,7 @@ use zeromq::*;
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Start server");
     let mut socket = zeromq::RepSocket::new();
-    socket.bind("127.0.0.1:5555").await?;
+    socket.bind("tcp://127.0.0.1:5555").await?;
 
     loop {
         let mut repl: String = socket.recv().await?.try_into()?;
--- a/examples/weather_client.rs
+++ b/examples/weather_client.rs
@@ -6,7 +6,7 @@ use zeromq::{BlockingRecv, Socket};
 async fn main() -> Result<(), Box<dyn Error>> {
     let mut socket = zeromq::SubSocket::new();
     socket
-        .connect("127.0.0.1:5556")
+        .connect("tcp://127.0.0.1:5556")
         .await
         .expect("Failed to connect");
 
--- a/examples/weather_server.rs
+++ b/examples/weather_server.rs
@@ -7,7 +7,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let mut rng = rand::thread_rng();
     println!("Start server");
     let mut socket = zeromq::PubSocket::new();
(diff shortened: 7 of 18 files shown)


TheButlah:
Update: I've removed `TryIntoEndpoint` and the associated `impl TryIntoEndpoint` parameters. I think making people do `"tcp://example.com:4567".parse()?` instead of just the string is not that unreasonable - and it makes the function signatures a lot simpler.

examples/message_broker.rs
```diff
@@ -7,11 +7,11 @@ use zeromq::{Socket, ZmqError};
 async fn main() -> Result<(), Box<dyn Error>> {
     let mut frontend = zeromq::RouterSocket::new();
     frontend
-        .bind("127.0.0.1:5559")
+        .bind("tcp://127.0.0.1:5559".parse()?)
```
Alexei-Kornienko, line 10: Imho this is not the best solution. I try to keep public api simple as possible and here you force user to make 2 error checks instead of one. bind() itself may return an error and IMHO we should just accept &str and do the parsing inside. In such case we can return appropriate error from bind call in case of malformed address

Alexei-Kornienko:
I think it's too early to declare support for all these transport types. IMHO we should start working on at least 1 additional transport and make sure that we can actually support it without big changes in code structure. So imho we should try and implement udp/ipc for example and when we'll be sure that it actually works do all the needed stuff to provide configuration options

Alexei-Kornienko (changes requested):
pls see my comments. Currently I would prefer not to merge this

Alexei-Kornienko:
For question 1. We already have a Address error variant https://github.com/zeromq/zmq.rs/blob/master/src/error.rs#L8 IMHO all this errors relate to this class

examples/message_broker.rs
TheButlah: Would the impl Trait approach solve that? See https://github.com/zeromq/zmq.rs/pull/68/commits/15a3c814e248c8b06f3bb5c8fae6228ce9effa63. In that version, you can simply provide "tcp://127.0.0.1:5559" without a parse or `?` operator

TheButlah:
> For question 1. We already have a Address error variant https://github.com/zeromq/zmq.rs/blob/master/src/error.rs#L8 IMHO all this errors relate to this class

@Alexei-Kornienko Would you be open to me submitting a separate PR to rename `ZmqError:Address` to `ZmqError::Endpoint`? That would be more consistent with zmq's distinction between endpoints and addresses.

TheButlah:
> I think it's too early to declare support for all these transport types. IMHO we should start working on at least 1 additional transport and make sure that we can actually support it without big changes in code structure. So imho we should try and implement udp/ipc for example and when we'll be sure that it actually works do all the needed stuff to provide configuration options

Fair enough. I marked the `Endpoint` enum as well as the `Transport` enums as `#[non_exhaustive]`, so I can reduce them down to the minimal set of transports we want to support as a short term goal, and we can add to those enums over time as we grow to support more.

Should I have only the currently supported transport protocal (in this case just tcp) in those enums, or should I add for example Udp too in preparation for adding support for Udp?

examples/message_broker.rs
Alexei-Kornienko: Yes I think that would be a better option for now

TheButlah: I just added back in the `impl Trait` approach. Does that resolve this issue?

Alexei-Kornienko:
> Fair enough. I marked the Endpoint enum as well as the Transport enums as #[non_exhaustive], so I can reduce them down to the minimal set of transports we want to support as a short term goal, and we can add to those enums over time as we grow to support more.

Yes

>Should I have only the currently supported transport protocal (in this case just tcp) in those enums, or should I add for example Udp too in preparation for adding support for Udp?

I would start with TCP only and add other transport types together with respective implementation

TheButlah:
Ok, I've removed all the transport variants except Tcp.

TheButlah:
OK, I ended up doing the error handling changes in this PR because it integrated too closely with the `Endpoint` enum.

I couldn't trivially use the existing `ZmqError::Address` because it can only be constructed from an `AddrParseError`, which cannot be manually constructed. Therefore I ended up making an enum specifically for errors when parsing endpoints, and I made the main ZmqError use `#[from] EndpointError` on a variant called `Endpoint`. I'm hoping that this will have the desired granularity and specificity, but still allow for easy aggregation of errors into a single ZmqError type.

Alternatively, I could modify the PR to get rid of  `EndpointError` and put the different variants into the `ZmqError` instead. I personally prefer the current style, but I'd like to know your thoughts.

Alexei-Kornienko:
LGTM. Could you please rebase your branch before merging? IMHO you could squash some commits to make history clean and nice

Alexei-Kornienko (approved):

TheButlah:
@Alexei-Kornienko OK, I've rebased and fixed up some of the commits to distill it down to the two main contributions. Merge when ready

PR #70, opened 2020-09-29
Allow endpoints with port 0 or unspecified interface

TheButlah:
Zeromq lets you bind to the unspecified ipv4 interface or the unspecified ipv6 interface. It also lets you bind to any available port via port 0. This PR adds the necessary logic for these cases, as well as unit tests for the functionality in `Pub` ports, along with the `Socket::binds()` function which returns a list of the resolved endpoints.

Side note: A lot of these tests are brittle to maintain, because we have to hard code the port numbers that won't clash with any other running processes, or any of the other ports in the unit/integration tests. Its probably a good idea to make a utility for tests to get port numbers that won't conflict with the other tests, but thats out of scope for this PR

### DIFF
--- a/src/dealer_router.rs
+++ b/src/dealer_router.rs
@@ -96,13 +96,16 @@ impl Socket for RouterSocket {
         }
     }
 
-    async fn bind(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<()> {
-        let mut endpoint = endpoint.try_into()?;
-        let stop_handle =
-            util::start_accepting_connections(&mut endpoint, self.backend.clone()).await?;
+    async fn bind(
+        &mut self,
+        endpoint: impl TryIntoEndpoint + 'async_trait,
+    ) -> ZmqResult<&Endpoint> {
+        let endpoint = endpoint.try_into()?;
+        let (endpoint, stop_handle) =
+            util::start_accepting_connections(endpoint, self.backend.clone()).await?;
         self._accept_close_handle = Some(stop_handle);
         self.binds.push(endpoint);
-        Ok(())
+        Ok(self.binds.last().unwrap())
     }
 
     async fn connect(&mut self, _endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<()> {
--- a/src/dealer_router.rs
+++ b/src/dealer_router.rs
@@ -9,7 +9,7 @@ use tokio::net::TcpStream;
 use tokio_util::codec::Framed;
 
 use crate::codec::*;
-use crate::endpoint::TryIntoEndpoint;
+use crate::endpoint::{Endpoint, TryIntoEndpoint};
 use crate::error::*;
 use crate::message::*;
 use crate::util::*;
@@ -75,6 +75,7 @@ impl SocketBackend for RouterSocketBackend {
 pub struct RouterSocket {
     backend: Arc<RouterSocketBackend>,
     _accept_close_handle: Option<oneshot::Sender<bool>>,
+    binds: Vec<Endpoint>,
 }
 
 impl Drop for RouterSocket {
@@ -91,19 +92,25 @@ impl Socket for RouterSocket {
                 peers: Arc::new(DashMap::new()),
             }),
             _accept_close_handle: None,
+            binds: Vec::new(),
         }
     }
 
     async fn bind(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<()> {
-        let endpoint = endpoint.try_into()?;
+        let mut endpoint = endpoint.try_into()?;
         let stop_handle =
-            util::start_accepting_connections(&endpoint, self.backend.clone()).await?;
+            util::start_accepting_connections(&mut endpoint, self.backend.clone()).await?;
         self._accept_close_handle = Some(stop_handle);
+        self.binds.push(endpoint);
         Ok(())
     }
 
     async fn connect(&mut self, _endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<()> {
         unimplemented!()
+    }
+
+    fn binds(&self) -> &[Endpoint] {
+        &self.binds
     }
 }
 
--- a/src/pub.rs
+++ b/src/pub.rs
@@ -170,26 +170,29 @@ impl Socket for PubSocket {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{Host, ZmqResult};
+    use crate::util::tests::{
+        test_bind_to_any_port_helper, test_bind_to_unspecified_interface_helper,
+    };
+    use crate::ZmqResult;
+    use std::net::IpAddr;
+
     #[tokio::test]
-    async fn test_bind_to_any() -> ZmqResult<()> {
-        let mut s = PubSocket::new();
-        assert!(s.binds().is_empty());
-        for _ in 0..4 {
-            s.bind("tcp://localhost:0").await?;
-        }
+    async fn test_bind_to_any_port() -> ZmqResult<()> {
+        let s = PubSocket::new();
+        test_bind_to_any_port_helper(s).await
+    }
 
-        let bound_to = s.binds();
-        assert_eq!(bound_to.len(), 4);
-        let mut port_set = std::collections::HashSet::new();
-        for b in bound_to {
-            let Endpoint::Tcp(host, port) = b;
-            assert_eq!(host, &Host::Domain("localhost".to_string()));
-            assert_ne!(*port, 0);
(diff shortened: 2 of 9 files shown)


src/pub.rs
```diff
@@ -105,6 +105,7 @@ impl MultiPeer for PubSocketBackend {
 
 pub struct PubSocket {
     pub(crate) backend: Arc<PubSocketBackend>,
+    binds: Vec<Endpoint>,
```
TheButlah, line 108: Let me know if the right location for this is in `TSocket`, rather than `TSocketBackend`. The actual `binds()` function should probably be in `Socket` either way, as it ought to be externally visible.

src/pub.rs
```diff
@@ -158,4 +161,38 @@ impl Socket for PubSocket {
...
+    fn binds(&self) -> &[Endpoint] {
+        &self.binds
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::util::tests::{
+        test_bind_to_any_port_helper, test_bind_to_unspecified_interface_helper,
```
TheButlah, line 174: `test_bind_to_any_port_helper` and the interface helper are only used by test cases, but they will be used by many different socket types. Let me know if this approach of putting them in the util::tests module is idiomatic.

src/util.rs
```diff
@@ -194,15 +194,19 @@ pub(crate) async fn peer_connected(socket: tokio::net::TcpStream, backend: Arc<d
...
-/// connections on it Returns stop_handle channel that can be used to stop
-/// accepting new connections
+/// connections on it.
+///
+/// Returns stop_handle channel that can be used to stop accepting new
+/// connections. Also mutates `endpoint` to reflect the resolved endpoint
+/// address.
 pub(crate) async fn start_accepting_connections(
-    endpoint: &Endpoint,
+    endpoint: &mut Endpoint,
```
TheButlah, line 203: I was a bit conflicted on how to communicate the resolved endpoint back to the caller. I could take an owned `endpoint` but then it would waste resources if this function errored (i.e. the caller might have to defensively clone the endpoint in case of failure if they wanted to try again). I could alternatively not mutate the endpoint at all, but rather return just the port number that was resolved.

src/util.rs
```diff
@@ -218,6 +222,16 @@ pub(crate) async fn start_accepting_connections(
...
     });
+    debug_assert_ne!(resolved_addr.port(), 0);
+    *port = resolved_addr.port();
+    let resolved_host: Host = resolved_addr.ip().into();
+    if let Host::Ipv4(ip) = host {
+        debug_assert!(ip == &mut resolved_addr.ip());
+        *host = resolved_host;
+    } else if let Host::Ipv6(ip) = host {
+        debug_assert!(ip == &mut resolved_addr.ip());
+        *host = resolved_host;
```
TheButlah, line 233: I think there are ipv6 addresses that map to ipv4 addresses and I wasn't sure if `local_addr` might convert the ipv6 one to ipv4 in those cases. If it did, the `Host` variant would change from ipv6 to ipv4 in those instances. I am not sure if `local_addr` actually does that transformation, or if its a bug or feature.

tests/pub_sub.rs
```diff
@@ -62,10 +62,10 @@ async fn test_pub_sub_sockets() {
         server_stop_sender.send(()).unwrap();
     }
     let addrs = vec![
+        "tcp://localhost:5553",
         "tcp://127.0.0.1:5554",
         "tcp://[::1]:5555",
         "tcp://127.0.0.1:5556",
-        "tcp://localhost:5557",
```
TheButlah, line 68: This caused flaky tests, as port 5557 was used by another test so it would non deterministically clash on port number

src/util.rs
```diff
@@ -218,6 +222,16 @@ pub(crate) async fn start_accepting_connections(
             }
         }
     });
+    debug_assert_ne!(resolved_addr.port(), 0);
+    *port = resolved_addr.port();
+    let resolved_host: Host = resolved_addr.ip().into();
+    if let Host::Ipv4(ip) = host {
+        debug_assert!(ip == &mut resolved_addr.ip());
```
TheButlah, line 229: I was surprised to find that when given the unspecified address (0.0.0.0 or ::), `local_addr()` did not resolve it to a particular ip addr like `127.0.0.1`. Is that expected behavior?

Alexei-Kornienko:
I like the idea of the request however I find implementation a little bit confusing. I don't like that we pass endpoint by reference (why can't we pass it by value?) and even more I don't like the idea that we pass it as mutable reference to use it as a return value.
```
util::start_accepting_connections(&mut endpoint, self.backend.clone()).await?;
```
I also don't understand how end used is supposed to use binds() method.

Maybe we should consider making Endpoint a return value of a successful bind call? In such case user will immediately receive appropriate endpoint and use it if needed?

src/dealer_router.rs
```diff
@@ -91,20 +92,26 @@ impl Socket for RouterSocket {
...
+            binds: Vec::new(),
         }
     }
 
     async fn bind(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<()> {
-        let endpoint = endpoint.try_into()?;
+        let mut endpoint = endpoint.try_into()?;
         let stop_handle =
-            util::start_accepting_connections(&endpoint, self.backend.clone()).await?;
+            util::start_accepting_connections(&mut endpoint, self.backend.clone()).await?;
```
Alexei-Kornienko, line 102: I would consider changing the API the following way:
```
async fn bind(&mut self, endpoint: &str) -> ZmqResult<Endpoint>
```
and parse endpoint further down in utils method. making it a return value of the `start_accepting_connections` call

Alexei-Kornienko (changes requested):

TheButlah:
OK, I have revised the API slightly. What are your thoughts?

src/dealer_router.rs
```diff
@@ -91,20 +92,29 @@ impl Socket for RouterSocket {
...
 
-    async fn bind(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<()> {
+    async fn bind(
+        &mut self,
+        endpoint: impl TryIntoEndpoint + 'async_trait,
+    ) -> ZmqResult<&Endpoint> {
         let endpoint = endpoint.try_into()?;
-        let stop_handle =
-            util::start_accepting_connections(&endpoint, self.backend.clone()).await?;
+        let (endpoint, stop_handle) =
```
Alexei-Kornienko, line 104: LGTM

TheButlah:
@Alexei-Kornienko ok, ready to merge when its stamped

Alexei-Kornienko:
@TheButlah reading your MR I realized that we have another issue that is related to this code part - https://github.com/zeromq/zmq.rs/issues/71 Basically we need to store multiple stop_handles somewhere for each bind call to stop accepting connections properly. Talking this into account it makes more sense to make binds: Vec<Endpoint>, a HashMap<Endpoint, oneshot::Sender<bool>> to store this stop handles. This 2 changes together will fit nicely into the code and solve both issues

TheButlah:
>Talking this into account it makes more sense to make binds: Vec<Endpoint>, a HashMap<Endpoint, oneshot::Sender>` to store this stop handles.

Attempting to do this has turned into a big issue - if I have `PubSocket.binds` be a `HashMap<Endpoint, oneshot::Sender>`, there is a question of how to permit the user to call `binds()` - binds can no longer be a slice, but rather will need to be an iterator. 

This is turning out to be rather tricky, and its compounded by the fact that I have a followup merge request that diverges from several of the signatures to a fair degree. If its ok with you, could we consider fixing the stop_handles to be out of the scope of this MR so I may get this one through, show you the follow up one, and then we can circle back to the stop handles?

Alexei-Kornienko:
@TheButlah sure we can do this in separate MRs

Alexei-Kornienko (approved):

src/dealer_router.rs
```diff
@@ -91,20 +92,29 @@ impl Socket for RouterSocket {
...
+            util::start_accepting_connections(endpoint, self.backend.clone()).await?;
         self._accept_close_handle = Some(stop_handle);
-        Ok(())
+        self.binds.push(endpoint);
```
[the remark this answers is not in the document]
TheButlah: While it's true that the caller could store the endpoint, I think it's important to be able to re-aquire the endpoint from just the socket. Suppose for example you have written your code to pass an already bound socket to some other module. If you later decide that you need to know the endpoint of the socket, you wouldn't have an easy way of getting it without bundling the two up wherever you want to use the socket.

Its definitely not critical as other workarounds exist, but it's helpful to be able to re-aquire it at any time

PR #72, opened 2020-10-04
Added ability to generically convert to &Endpoint without needing ownership of value

TheButlah:
Rebased on #76, suggest reviewing that one first

TryIntoEndpoint has a signature of `try_into(self)`, which takes ownership over `Self`. This is not conducive to situations where we only have a `&Endpoint` or only need a `&Endpoint` returned, such as in `Socket::connect()`. This PR adds `TryAsRefEndpoint`, a failable version of `AsRef<Endpoint>`. This allows generically passing either `&Endpoint` or `&str` values to a function, and getting out `&Endpoint`, without needing to move in the endpoint or perform a clone.  This allows functions that would want to take `&Endpoint` and return `&Endpoint` to be generic.

~In addition, this PR breaks up the endpoints module into submodules for readability, fixes a bug associated with parsing and displaying `Host`s, and fixes a flaky test.~

### DIFF
--- a/src/dealer_router.rs
+++ b/src/dealer_router.rs
@@ -9,7 +9,7 @@ use tokio::net::TcpStream;
 use tokio_util::codec::Framed;
 
 use crate::codec::*;
-use crate::endpoint::{Endpoint, TryIntoEndpoint};
+use crate::endpoint::{Endpoint, TryAsRefEndpoint, TryIntoEndpoint};
 use crate::error::*;
 use crate::message::*;
 use crate::util::*;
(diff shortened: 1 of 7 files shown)


TheButlah:
Just rebased on top of #76, that should greatly reduce the scope of this PR

TheButlah:
I'm going to close this - as we discussed elsewhere, I'm not sure that avoiding a copy of the endpoint is as important for performance as I originally thought - binding and connecting are not in inner loops

PR #75, opened 2020-10-04
Refactored binds to HashMap. Fixes #71

### LINKED ISSUE #71: Multiple calls to bind will cause resource leak
User may call `socket.bind("...")` multiple times. Overall socket should work properly (accepting connections from several endpoints) However only last stop_handle will be used to stop accepting new connections. Handles from previous call will be lost and cause possible issues/resource leak

IMHO we should either prevent user from calling bind multiple times or make sure that we save all stop handles in some vec

Alexei-Kornienko:


Alexei-Kornienko:
@TheButlah pls review ^

src/rep.rs
```diff
@@ -25,11 +26,10 @@ struct RepSocketBackend {
 
 pub struct RepSocket {
     backend: Arc<RepSocketBackend>,
-    _accept_close_handle: Option<oneshot::Sender<bool>>,
     _fair_queue_close_handle: oneshot::Sender<bool>,
     current_request: Option<PeerIdentity>,
     fair_queue: mpsc::Receiver<(PeerIdentity, Message)>,
-    binds: Vec<Endpoint>,
+    binds: HashMap<Endpoint, oneshot::Sender<bool>>,
```
TheButlah, line 32: Should the `_fair_queue_close_handle` be in here too? What does that handle do?

src/lib.rs
```diff
@@ -135,15 +136,13 @@ pub trait Socket {
...
-        -> ZmqResult<&Endpoint>;
+    async fn bind(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<Endpoint>;
 
     // TODO: Although it would reduce how convenient the function is, taking an
     // `&Endpoint` would be better for performance here
     async fn connect(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<()>;
 
-    /// Retrieves the list of currently bound endpoints
-    fn binds(&self) -> &[Endpoint];
+    fn binds(&self) -> &HashMap<Endpoint, oneshot::Sender<bool>>;
```
TheButlah, line 145: My concern here is that we are exposing in the public api the implementation details. Ideally this would be an iterator of some sort. It also forces us to have one representation of the hashmap for all the different socket types - how do we know that a different socket type might not need a value of a different type, unique to that socket type? Customizing the type of the hashmap or the iterator is probably impossible without generic associated types, so this may become a real issue later. 

Its definitely not a blocker for this MR as its more important to fix the resource leak - I can try to address it in a followup PR

src/dealer_router.rs
```diff
@@ -74,8 +75,7 @@ impl SocketBackend for RouterSocketBackend {
 
 pub struct RouterSocket {
     backend: Arc<RouterSocketBackend>,
-    _accept_close_handle: Option<oneshot::Sender<bool>>,
-    binds: Vec<Endpoint>,
+    binds: HashMap<Endpoint, oneshot::Sender<bool>>,
 }
 
 impl Drop for RouterSocket {
```
TheButlah, line 81: Do we need to signal the channel on shutdown? If not, could you add a oneline comment explaining why not? Also as a probably unrelated question, does backend.shutdown() do anything that simply dropping the backend doesnt? Is there a reason you would want to shut the backend down without dropping it?

src/lib.rs
```diff
@@ -135,15 +136,13 @@ pub trait Socket {
     /// connections on it.
     ///
     /// Returns the bound-to endpoint, with the port resolved (if applicable).
-    async fn bind(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait)
-        -> ZmqResult<&Endpoint>;
+    async fn bind(&mut self, endpoint: impl TryIntoEndpoint + 'async_trait) -> ZmqResult<Endpoint>;
```
TheButlah, line 139: I think in an ideal world, we would never need to clone the endpoint once we convert it, so in theory the endpoint returned should be a ref, rather than owned. That being said, internally we have to clone right now in order to insert into the hashmap. But ideally this would be unnecessary.

I can take a crack at avoiding the clone, and playing around with the return value, in the same PR where I take a crack at the `binds()` return type

TheButlah (commented):
Could you add a test case to ensure that all of the accept handlers get closed? That way we can be sure we don't accidentally re-introduce a resource leak in the future

src/rep.rs
Alexei-Kornienko: not really. these are two different logical things (even so the callback is the same type). binds callback is responsible for closing couroutine that is accepting new connections. completely separate from that is another couroutine that is responsible for merging messages from all incoming connections into a single queue (according to fair queue rules)

src/lib.rs
Alexei-Kornienko: I agree that this is not the best option and we may need to rework it later. For now I've just implemented a quick-fix that solves 1 particular issue without big changes in other parts.

src/dealer_router.rs
Alexei-Kornienko: > Also as a probably unrelated question, does backend.shutdown() do anything that simply dropping the backend doesnt? Is there a reason you would want to shut the backend down without dropping it?

Droping backend will not stop everything properly. Since backend is kept inside by an `Arc` other references may remain preventing it from closing. Main idea behind `shutdown` call is that it's built in such a way that it notifies other coroutines holding references to a backend that they should stop working.

> Do we need to signal the channel on shutdown? If not, could you add a oneline comment explaining why not?

Basically oneshot::Sender is built in such a way that it notifies a Receiver part when it's dropped. So there is no need to signal it explicitly.

src/lib.rs
Alexei-Kornienko: IMHO this is not the most important part to work on. Basically bind/connect is called only a several times comparable to the number of call to send/recv because of this additional cloning has negligible impact on performance cause it's definitely not on a critical code path. We may come back to this some time later, however for now I mostly focus on avoid additional cloning of message contents for example. For other parts I plan to return later after actual benchmarking and finding real issues.

TheButlah (approved):
LGTM - we will circle back to the signature of `binds()` later. Adding a test case to make sure we kill the coroutine would be really great, its up to you if you want to do that in this PR

PR #78, opened 2020-10-11
Ported code to use futures_codec and tokio_util's Compat

TheButlah:
This MR is a smaller change than #77, which avoids any dynamic dispatch or boxing. It is intended to isolate the changes to just that what is necessary to use `futures_codec` and `tokio_util::compat` instead of `tokio_util::codec`. 

~**Don't merge this MR** - we need to merge some basic performance benchmarks to master first,~ and then we can see the performance impact of the codec changes, separately from the dynamic dispatch stuff in #77.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,9 +29,13 @@ futures_codec = "0.4"
 chrono = "^0.4"
 criterion = "0.3"
 
-# [[bench]]
-# name = "pub_sub"
-# harness = false
+[lib]
+bench = false
+
+[[bench]]
+name = "pub_sub" 
+harness = false 
+bench = false # Don't actually benchmark this, until we fix it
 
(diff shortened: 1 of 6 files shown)


TheButlah:
OK, using criterion's CI, I ran a benchmark for both master and for this PR. both were run with a warm up time of 120 and a measurement time of 60 seconds. Saving both results and comparing the two, here's what I get:
image

I trust these results, the measurement and warm ups were long enough and my clock speeds stayed the same in both after the warm up period. So it looks like we have a 3.4% performance regression.

Personally I think that's totally fine given the benefits since we can probably optimize stuff later, but what do you think @Alexei-Kornienko?

Alexei-Kornienko:
LGTM. I don't see any noticeable regression on my PC

PR #79, opened 2020-10-11
Broke up codecs module and added CodecError

TheButlah:
Largely a refactor. The codecs module was getting a bit monolothic and hard to work with, so I wanted to break it up into submodules. I also introduced a CodecError type, which is supposed to be an error when converting from bytes to frames in the codec.This should help the specificity when dealing with codec errors. I also introduced `FramedIo<Io>` which is just a newtype on `tokio_util::codec::Framed<Io, ZmqCodec>`


Alexei-Kornienko:
LGTM. will merge after rebase

TheButlah:
@Alexei-Kornienko done

PR #83, opened 2020-10-13
Further generalize across transports

TheButlah:
~Rebased on #77 - I suggest diffing against that for review~

In preparation for new transport types, this PR takes steps to generalize the signatures of `transport::connect()` and `transport::begin_accept()`. This should help make it trivial to add support for new transports to pubsub and reqrep. It also wraps the oneshot channel for killing the accept task in a struct to increase type safety.

### DIFF
--- a/src/endpoint/mod.rs
+++ b/src/endpoint/mod.rs
@@ -7,6 +7,7 @@ pub use transport::Transport;
 use lazy_static::lazy_static;
 use regex::Regex;
 use std::fmt;
+use std::net::SocketAddr;
 use std::str::FromStr;
 
 use crate::error::EndpointError;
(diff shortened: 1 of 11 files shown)


src/endpoint/mod.rs
```diff
@@ -38,6 +39,11 @@ impl Endpoint {
             Self::Tcp(_, _) => Transport::Tcp,
         }
     }
+
+    /// Creates an `Endpoint::Tcp` from a [`SocketAddr`]
+    pub fn from_tcp_sock_addr(addr: SocketAddr) -> Self {
```
Alexei-Kornienko, line 44: why not implementing `From<SocketAddr> for Endpoint`?

Alexei-Kornienko (commented):
One minor question. Otherwise LGTM

src/endpoint/mod.rs
TheButlah: `SocketAddr` could be either TCP or UDP, so its not clear which transport it should be without making it explicit

Alexei-Kornienko: Ah, I see...

Issue #87, opened 2020-10-14
Ipc sockets have race conditions when binding to the same path


TheButlah:
When socket `B` binds to a path used by socket `A`, `B` gets an error. However, if `A` is deleted/unbound and then `B` immediately attempts to bind, the path may not yet have been deleted. In this case, `B` gets an error when the code would imply that the socket path should no longer be used by `A`.

We should probably add some retry logic in the bind method of ipc sockets (or all sockets?) to alleviate this issue, or otherwise we should have an async unbind method that doesn't return until the file is deleted and unbound. I suspect this latter approach is the ideal one

TheButlah:
@Alexei-Kornienko I suspect that the reason this is happening is that when the socket is `drop`ped, the `AcceptStopChannel` gets dropped, which causes the accept loop to break. The cleanup code in the accept task will then run, but nothing prevents the `drop()` method of the socket from blocking until that completes. Instead, we should have an `unbind()` method that sockets can call to unbind. This will be invoked in the `drop()` function.

In order for such an unbind function to work, we need to hold the JoinHandle of the accept task, in addition to the `AcceptStopChannel`. Where should I store this data? I am not sure if it should go in the socket backend or the regular socket struct. I'm still a bit unclear on what the difference between the two is.

Alexei-Kornienko:
We won't be able to call `unbind()` in `drop()` function. unbind would be async and drop must be synchronous. Currently it works in such a way that dropping socket will `eventually` free all allocated resources.

TheButlah:
You're right - I just learned this the hard way after going down the rabbit hole of Async Drop RFCs and other blog posts.

What I'm currently looking at is introducing a general purpose `TaskHandle<T>` struct, that has one method- `async fn shutdown(self) -> ZmqResult<T>`. The task handle will contain both the shutdown channel as well as the JoinHandle from tokio. We can't run `shutdown()` in the destructor (due to the aforementioned issue), but we can hold this handle and call shutdown on it where appropriate in our code where we need to await on the task. In the general case where we instead `Drop` the handle without any explicit async shutdown, we still know that it will eventually shutdown, because the stop channel is cancelled and the task will then leave its `loop`.

This strikes a good balance to me - eventual shutdown as a default, with the ability to await shutdown when needed.

TheButlah:
This is fixed by #89

PR #88, opened 2020-10-14
Improved error msgs, added TaskHandle, TaskError, AcceptStopHandle

TheButlah:
Adds a few primitives that should make it easier to work with tasks going forward. Also refactors the code to use this for the accept handle.

Does not yet address #87

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ log = "0.4"
 futures_codec = "0.4"
 
 [dev-dependencies]
-chrono = "^0.4"
+chrono = "0.4"
 criterion = "0.3"
 pretty_env_logger = "0.4"
 
--- a/src/task_handle.rs
+++ b/src/task_handle.rs
@@ -1,0 +1,61 @@
+pub use self::tk::TaskHandle;
+
+use crate::error::{ZmqError, ZmqResult};
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum TaskError {
+    #[error("Internal task error: {0}")]
+    Internal(#[from] Box<ZmqError>),
+    #[error("Task panicked")]
+    Panic,
+    #[error("Task cancelled")]
+    Cancelled,
+}
+impl From<tokio::task::JoinError> for TaskError {
+    fn from(err: tokio::task::JoinError) -> Self {
+        if err.is_panic() {
+            TaskError::Panic
+        } else {
+            debug_assert!(err.is_cancelled());
+            TaskError::Cancelled
+        }
+    }
+}
+
+mod tk {
+    use super::*;
+
+    pub struct TaskHandle<T> {
+        // Using options to allow us to move resource without consuming `self`
+        stop_channel: futures::channel::oneshot::Sender<()>,
+        join_handle: tokio::task::JoinHandle<ZmqResult<T>>,
+    }
+    impl<T> TaskHandle<T> {
+        pub(crate) fn new(
+            stop_channel: futures::channel::oneshot::Sender<()>,
+            join_handle: tokio::task::JoinHandle<ZmqResult<T>>,
+        ) -> Self {
+            Self {
+                stop_channel,
+                join_handle,
+            }
+        }
+
+        /// Shutdown the task and return the task's result, consuming the handle
+        /// in the process
+        #[allow(dead_code)]
+        pub(crate) async fn shutdown(self) -> ZmqResult<T> {
+            // Its ok to ignore possible error, because a dropped channel it has the same
+            // effect as sending stop signal
+            let _ = self.stop_channel.send(());
+            let join_result = self.join_handle.await.map_err(TaskError::from);
+            match join_result {
+                Ok(Ok(ok)) => Ok(ok),
+                Ok(Err(zmq_err)) => Err(zmq_err),
+                Err(task_err) => Err(task_err.into()),
+            }
+        }
+    }
+}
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ log = "0.4"
 futures_codec = "0.4"
 
 [dev-dependencies]
-chrono = "^0.4"
+chrono = "0.4"
```
Alexei-Kornienko, line 29: Why do you pin it to certain release? I thought that using ^0.4 notation is semantically better cause it will automatically use newer version when available.

src/task_handle.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+}
+
+mod tk {
+    use super::*;
+
+    pub struct TaskHandle<T> {
+        // Using options to allow us to move resource without consuming `self`
+        stop_channel: Option<futures::channel::oneshot::Sender<()>>,
```
Alexei-Kornienko, line 32: What's the point in wrapping this into Option? Can we ever construct/use this struct with None values?

Cargo.toml
TheButlah: Caret requirements are equivalent to requirements without the caret symbol. From the cargo guide:
>The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

From my brief survey of other open source libraries, I saw that it seemed to be the established convention to prefer the non-caret syntax, which is why I personally specify things with the non-caret sytax. However, both are identical - I'm good to do the caret one if you feel strongly on that. 
Sources cited: [rocket](https://github.com/SergioBenitez/Rocket/blob/master/core/lib/Cargo.toml), [thiserror](https://github.com/dtolnay/thiserror/blob/master/Cargo.toml), [actix](https://github.com/actix/actix/blob/master/Cargo.toml), [tokio](https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml), [log](https://github.com/rust-lang/log/blob/master/Cargo.toml), [rand](https://github.com/rust-random/rand/blob/master/Cargo.toml), [syn](https://github.com/dtolnay/syn/blob/master/Cargo.toml), [lazy_static](https://github.com/rust-lang-nursery/lazy-static.rs/blob/master/Cargo.toml), [futures](https://github.com/rust-lang/futures-rs/blob/master/futures/Cargo.toml), etc (I didn't find any examples with the caret symbol that I looked at)

src/task_handle.rs
TheButlah: Originally I did this because I needed the shutdown function to take an `&mut Self`, but it takes a `Self` now. I can probably change these back to the option-less version.

Cargo.toml
Alexei-Kornienko: > Caret requirements are equivalent to requirements without the caret symbol.

It seems that I missed that part of the book..

src/task_handle.rs
Alexei-Kornienko: Imho it's better not to use Option when it's not needed. This makes this structure more reliable

TheButlah: Oh I remember now...

```
error[E0509]: cannot move out of type `TaskHandle<T>`, which implements the `Drop` trait
  --> src/task_handle.rs:54:32
   |
54 |             let stop_channel = self.stop_channel; //.take().unwrap();
   |                                ^^^^^^^^^^^^^^^^^
   |                                |
   |                                cannot move out of here
   |                                move occurs because `self.stop_channel` has type `futures::channel::oneshot::Sender<()>`, which does not implement the `Copy` trait
   |                                help: consider borrowing here: `&self.stop_channel`
```

I implemented Drop so that we can get warnings when we dropped the contents during panics without shutting down. For that reason, I can't actually destructure the struct in a way that lets me consume its fields. To get around it, I used `Option` to allow me to `take()` the contents and assume ownership over them.

Alexei-Kornienko: Do we really need this log?
Even in case of panic stop_channel will be dropped and this will cause task to stop (just without explicit wait).

TheButlah: So we have `shutdown()` which we can call whenever we need an explicit way to block until the task is finished - in the case of the `ipc` socket logic for example, shutdown is important to make sure resources are released before doing further operations to avoid a race condition that would leak a file descriptor and prevent rebinding to the same socket file. 

However, suppose we panic before calling `shutdown()`. In this case, the task will eventually cancel due to its `drop()` implementation, but we have potentially introduced a race condition somewhere else in the code (since the parent task was supposed to `await` until the child task completed). For debugging purposes therefore, it was helpful for me when I was doing to IPC stuff to see if we ever drop a `TaskHandle` while panicking.

Alexei-Kornienko: > However, suppose we panic before calling shutdown(). In this case, the task will eventually cancel due to its drop() implementation, but we have potentially introduced a race condition somewhere else in the code (since the parent task was supposed to await until the child task completed).

From what I know we are not expected to provide any kind of guarantees in case of panic. In most cases this will simply stop current process and exit.

I don't think that it's a good idea to change code structure just for debug purposes (cause it introduces some additional complexity without any good reason). Without Option you could also remove is_shutting_down property (cause you can only shutdown this Handle once). Overall it would make it simpler and easier to understand.

TheButlah:
Ok, I can do that

TheButlah:
OK, I believe I have resolved the issue. Let me know if its good to merge

PR #90, opened 2020-10-15
Added ipc benchmark

TheButlah:
Adds ipc to the benchmarks for each socket type, and groups them against the other transports for comparison

### DIFF
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -3,14 +3,20 @@ use tokio::runtime::Runtime;
 
 use zeromq::{prelude::*, PubSocket, SubSocket};
 
+type BenchGroup<'a> = criterion::BenchmarkGroup<'a, criterion::measurement::WallTime>;
+
 /// Binds n pubs and connects n subs to them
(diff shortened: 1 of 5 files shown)


TheButlah:
WSL2 results:
image
image
Surprised that the IPC is almost the same speed as tcp.

TheButlah:
Ubuntu 20 results:
image
image
Its interesting that the performance on Ubuntu is sooo much higher than in WSL2. I wonder if microsoft might be misleading us about just how performant WSL is... Either way, the difference in speed is a lot more noticeable now, so thats good. Not as big of a gap as I expected however.

Alexei-Kornienko:
LGTM. @TheButlah WSL is definitely slow.. however regarding the benchmark performance - it may happen that we are doing something pretty stupid regarding message processing inside the driver. This may lead to the fact that performance is similar cause maybe 70% of the time is consumed by internal handling and not message going through tcp/ipc. The next steps is to actuall profile code that is running benchmark. If you are interested in it you can try to start with https://github.com/jonhoo/inferno this is a pretty neat project that explains how to use perf to collect results and displays them in nice graphs

PR #92, opened 2020-10-16
Added pub_sub and req_rep compliance test

TheButlah:
Rebased on #89 

This hangs and doesn't actually seem to receive any messages from `their_pub`, although the handshape appears to complete properly. Something is afoot, so I thought I'd submit the PR as a draft so we can investigate.

Note that the pub_sub benchmark fails to receive also. Perhaps there is a connection?


Alexei-Kornienko:
Run it several times:

```
     Running target/debug/deps/pub_sub_compliant-cc8d0819d7576970

running 1 test
(ACCEPTED, 16, "tcp://127.0.0.1:38041")
(HANDSHAKE_SUCCEEDED, 0, "tcp://127.0.0.1:38041")
Finished pub task
(CLOSED, 12, "tcp://127.0.0.1:38041")
(DISCONNECTED, 16, "tcp://127.0.0.1:38041")
(MONITOR_STOPPED, 0, "")
test test_their_pub_our_sub ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```
Works for me on Ubuntu

Alexei-Kornienko:
@TheButlah  One of the reasons this might happen to you is because you are using a PUB/SUB sockets wrong. 

You cannot expect to send exactly 64 messages for example and receive all 64 of them. That's not the way how PUB works.

When pub socket doesn't have any subscribers it will simply throw messages away. Cause of this it may happen that PUB socket is open to soon and several messages gets lost. Then you try to receive a predefined number of messages via Sub socket and that never happens cause PUB is already finished at that time.

Instead you should write PUB part in such a way that it sends messages indefinetely (until it receives a signal). And SUB socket connects to listen some of them. That's why I used to send current time through PUB socket. Just to make sure that messages are unique. Imaging that PUB is for example NTP server that is broadcasting time to whoever wants to listen

TheButlah:
@Alexei-Kornienko you're right - I inadverdently resolved the issue via the time sleep command, which is why it was working. I've improved the robustness of the test now to account for the slow joiner problem and everything is working properly. Ready for final review.

Alexei-Kornienko:
LGTM

PR #95, opened 2020-11-05
Added Dealer socket impl. Minor fixes in Router socket as well

Alexei-Kornienko:


src/dealer.rs
```diff
@@ -1,12 +1,211 @@
...
+                    let send_result = peer.send_queue.try_send(messages);
+                    match send_result {
+                        Ok(()) => {
+                            self.backend.round_robin.push(next_peer_id.clone());
+                            return Ok(());
+                        }
+                        Err(e) => {
+                            if e.is_full() {
+                                // Try again later
+                                self.backend.round_robin.push(next_peer_id.clone());
```
Alexei-Kornienko, line 200: This may cause infinite loop. TODO - Think about how this should be avoided

PR #99, opened 2020-11-24
Alex/c bench

Alexei-Kornienko:
This MR adds a simplistic implementation of REQ/REP socket written in C using regular ZMQ driver.
Later I use this impl. to test and profile performance of Rust version.
Before this change Rust was ~40% slower than C impl.
After this change Rust is ~15 faster than C variant

### DIFF
--- a/src/rep.rs
+++ b/src/rep.rs
@@ -61,7 +61,7 @@ impl Socket for RepSocket {
 }
 
 impl MultiPeerBackend for RepSocketBackend {
-    fn peer_connected(&self, peer_id: &PeerIdentity, io: FramedIo) {
+    fn peer_connected(self: Arc<Self>, peer_id: &PeerIdentity, io: FramedIo) {
         let (recv_queue, send_queue) = io.into_parts();
 
         self.peers.insert(
@@ -81,10 +81,7 @@ impl MultiPeerBackend for RepSocketBackend {
     }
 }
 
-#[async_trait]
 impl SocketBackend for RepSocketBackend {
-    async fn message_received(&self, peer_id: &PeerIdentity, message: Message) {}
-
     fn socket_type(&self) -> SocketType {
         SocketType::REP
     }
--- a/src/rep.rs
+++ b/src/rep.rs
@@ -16,7 +16,7 @@ use std::sync::Arc;
 
 struct RepPeer {
     pub(crate) _identity: PeerIdentity,
-    pub(crate) send_queue: mpsc::Sender<Message>,
+    pub(crate) send_queue: FramedWrite<Box<dyn FrameableWrite>, ZmqCodec>,
     pub(crate) recv_queue_in: mpsc::Sender<Message>,
     pub(crate) _io_close_handle: futures::channel::oneshot::Sender<bool>,
 }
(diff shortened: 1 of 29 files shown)


src/rep.rs
```diff
@@ -75,51 +60,28 @@ impl Socket for RepSocket {
...
             },
         );
-        self.peer_queue_in
-            .clone()
-            .try_send((peer_id.clone(), in_queue_receiver))
-            .unwrap();
-
-        (out_queue_receiver, stop_callback)
+        self.fair_queue_inner
+            .lock()
```
TheButlah, line 75: Curious about the ramifications of this - I thought that since acquiring a mutex is a blocking operation, doing it in async code would result in all tasks on the current thread in the async executor being blocked? Is this actually performant when scaling to many async tasks at once? Perhaps its faster on one thread, but I am suspicious of how this scales

Alexei-Kornienko: Good question. In theory it may become an issue. However my thoughts are currently the following:
1) Contention happens always between 2 threads (one processing incoming connections, and the other is user thread actually receiving data). So contention doesn't become worse with more spawned tasks. I assume that in most cases it would just take an atomic operation or perhaps a spin lock to actually take this mutex.
2) Changes made by this MR were made after some benchmarking and comparing with C code doing equivalent stuff. After this changes performance improved and now it's even a bit better than C++ implementation. If you could provide some kind of test scenario when this becomes an issue we can think what we can do about it.
3) I didn't find a better way of doing this. Basically 2 threads need to communicate and share data. Sharing data via channels also has some internal synchronization. If you can propose a better way of doing this we can try to benchmark and discuss it.

TheButlah: What about using the async version? https://docs.rs/tokio/1.0.0/tokio/sync/struct.Mutex.html
I could submit a PR that switched to the async mutex and we can benchmark to see if there is a performance regression. Barring a regression, we could then switch to the async version "just in case"

Alexei-Kornienko: I've seen this mutex already, basically it's description says:
>The primary use case of the async mutex is to provide shared mutable access to IO resources such as a database connection. If the data stored behind the mutex is just data, it is often better to use a blocking mutex such as the one in the standard library or parking_lot. This is because the feature that the async mutex offers over the blocking mutex is that it is possible to keep the mutex locked across an .await point, which is rarely necessary for data.

I don't need to carry locked mutex over await so I decided that simple sync mutex should be sufficient.

>Can you elaborate? I didn't fully understand this

Efficiency and performance of a single mutex is determined by 2 main factors:
1) Length of critical sections (time while the mutex is locked). I tried to minimize this by using only a simplest operations while the mutex is locked.
2) Total number of threads trying to lock mutex concurrently. I was actually wrong claiming that only 2 thread content for the mutex (actually there are 3 of them - 1 thread processing incoming connections, 2 user thread, 3 async reactor thread that notifies future when new data received). Events generated by thread 1 (the code you are pointing to) only happen occasuly so it should not affect performance a lot. Main contention happens between thread 2 and 3.

From what I know about the implementation of parking lot mutex (and some other mutexes as well) It can work in 3 basic modes:
1) trying to lock an unlocked mutex - this happens very fast cause it's done by atomic operation
2) trying to lock a locked mutex - mutex uses atomic operation and spin lock for some number of iterations. It works well for short critical sections and low contention scenarios.
3) trying to lock a locked mutex - uses OS calls to suspend thread and schedule a wakeup call when mutex will be freed. This is a last resort variant if 2 of the above failed.

So my idea is that there is a high probability that this mutex will be locked with a fast path variant and we don't need to worry about blocking thread.
If it will not be the case we still need to block *some* thread to actually get the job done but it may happen that we'll block reactor thread and it will affect performance. This variant still needs to be tested and benchmarked. Currently I don't have any evidence that this actually happens and creates a bottleneck.

TheButlah: Interesting - point made. Not sure what the best way to actually check for bottlenecks is then, other than perhaps many sockets all running at once?

Alexei-Kornienko: Sure such scenario will make a good benchmark in any case. We just need to organize this bench in some reasonable way that would be close to real world scenario and would provide consistent and repeatable results

Alexei-Kornienko: One interesting thing to take into account:
While doing benchmarks and comparison with C/C++ variant (200K req/rep messages) I also took "Voluntary context switches" metric into account basically it gives some idea about how often your thread/process has nothing useful to work on and goes asleep (or goes asleep waiting for mutex).

C variant - Voluntary context switches: 600725
Rust variant before refactoring - Voluntary context switches: 911627
Rust variant after refactoring - Voluntary context switches: 399956

PR #100, opened 2020-11-28
Relax ZMTP version requirements according to the spec

malor:
One of the compliance tests was failing for me locally, because the peer was using a newer version of the ZMTP protocol.

According to the spec, instead of checking for a specific version supported by this implementation, we should allow any version that is equal or greater, as in this case the peer should adjust to us. If they choose not to, they are expected to close the connection immediately.

Similarly, this implementation may support communication with peers using the older version of the ZMTP protocol in the future.


Alexei-Kornienko:
LGTM

PR #106, opened 2020-12-25
Update to Tokio 1.0, Fix racey test

TheButlah:
Rebased on #107 due to CI not running

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,8 @@ async-trait = "0.1"
 parking_lot = "0.11"
 rand = "0.7"
 bytes = "0.5"
-tokio = { version = "0.2", features = ["full", "tcp"] }
-tokio-util = { version = "0.3", features = ["compat"] }
+tokio = { version = "1", features = ["full"] }
+tokio-util = { version = "0.6", features = ["compat"] }
 num-traits = "0.2"
 enum-primitive-derive = "0.1"
 dashmap = "3.11"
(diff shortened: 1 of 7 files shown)


Alexei-Kornienko:
Hmm CI failed... seems some tests are flacky again :( . I've seen it couple of times but didn't take the chance to actually debug and fix this issue. @TheButlah I'm to merge this as is. I don't see any harm in this MR

TheButlah:
Code currently fails to pass a test - the asserted socket event is different. I don't really know what the expected behavior is here, does the test need to be fixed or did updating to 1.0 somehow actually break things or change stuff?

Alexei-Kornienko:
@TheButlah does it fail for master on your machine? Or it happens only for this branch?

TheButlah:
Does not fail on master

Alexei-Kornienko:
Strange.. I've seen this error before. From what I understand it's related to order of events in C++ implementation of sockets. It seems that order of the events is not determined and sometimes you receive CLOSE event before last DISCONNECTED event.
IMHO the right way of fixing this would be to rethink the way we do assertions in this test and don't rely on order and number of events from C++ monitor queue

TheButlah:
OK thats fair - I did it this way because I simply observed this when running equivalent C++ code. But if the actual order isn't supposed to be determined, then instead I can keep a set of expected events, and do things that way (instead of absolute ordering)

TheButlah:
OK - its not so much that the C++ sockets themselves give nondeterministic events, but rather their publisher would get `drop`ped on receiving the last event. At the same time, our subscribers would `drop` when they sent the last events. So it wasn't guaranteed that their publisher would live long enough to observe the disconnection of our subs.

I've fixed this and submitted a commit in this branch.

Issue #64, opened 2020-09-23
Support WASM to allow people to use native zmq in the web browser


TheButlah:
One of Rust's strengths is is wasm support, letting developers run things natively in the web browser. It would be great if we could plan to support wasm in the future.

Looking into this, tokio currently [doesn't support wasm](https://github.com/tokio-rs/tokio/issues/1597), and async-std has a [minimal feature set that supports wasm](https://github.com/async-rs/async-std/issues/220). It seems like a trivial way to use tcp in the browser is not yet feasible, but we should be on the lookout for ways to support it in the near future. One possible approach might be to use websockets instead of tokio sockets, and perhaps unbake the assumption that the sockets are managed by tokio by abstracting them with a common trait.

TheButlah:
I've added `ws://` and `wss://` to the transport todo list here: #65. Unsure if thats the right approach, but we can figure it out in the future.

TheButlah:
Update: I've learned a bit more about web sockets vs tcp. Web sockets is a protocol, not a transport (i.e. HTTP vs tcp). It is built on top of TCP. However, you cannot connect a websocket client to a tcp socket unless the server is also using websockets - just like you can't make an HTTP request to a tcp socket unless its actually a HTTP server listening on that socket.

This is why the `ws://` and `wss://` transports exist - they descibe how ZMQ sockets work over web sockets. Therefore, you will never be able to do `tcp://` in the browser, but you can do `wss://` in the browser and connect to a server using `wss:://`.

TheButlah:
I've looked into the ecosystem to support this a bit.

First, there is the question of the async executor. Tokio does not support WASM, and doesn't seem to have any plans to do so. async-std currently has paritial support for WASM - importantly, it builds for the platform and task::spawn works (according to [here](https://github.com/async-rs/async-std/issues/220)). Its important to note that we technically care only about the *executor*, not the other functionality of tokio/async-std, but we may consider transitioning as an option. We should try to take care to write our code as runtime-independent as possible.

Next, there is the question of how to get a websocket library that works with the async ecosystem. The best option looks to be [async-tungstenite)(https://github.com/sdroege/async-tungstenite).

TheButlah:
@Alexei-Kornienko would you be open to a PR to get the code to work on both async-std and tokio, via a feature flag?

Alexei-Kornienko:
[...]

PR #108, opened 2020-12-25
Add async-std support

TheButlah:
Review #109 first - this is based on that.

To avoid contributing to the fragmentation of the async ecosystem, I think its really important to support both tokio and async-std. Additionally, async-std is the only runtime that has wasm support currently, and it is a personal goal of mine to get this project capable of running in the browser and communicating over web sockets (see #64).

You can control which version via feature flags, as this seemed to be the canonical approach in other projects that can use either runtime (like [async-tungstenite](https://github.com/sdroege/async-tungstenite)) for example. I've left the default runtime to be tokio, as it seems to be the more popular and fully featured option.

In terms of the complexity of maintaining two versions, the two libraries present an extremely similar api. In most places, the wrappers I wrote are a drop in replacement for whatever we were using before, and you can see that with just wrappers for spawning tasks. everything else is just in how transports are implemented, and even there the code doesn't diverge much. I had to add some beginner-unfriendly cfg statements in the example code, but that is a small price to pay for being able to build all tests and examples in either version, rather than just the source code. Also, both tokio and async-std are 1.0 now, so we shouldn't need to worry to much about changing APIs.

I also now test both tokio and async std in the CI. This should ensure that our code operates on both versions.


Alexei-Kornienko:
Mostly LGTM. Could you please add some description in README explaining how to properly build/run examples with different runtimes? IMHO examples become a little bit more complicated cause now they support both runtimes.

TheButlah:
@Alexei-Kornienko OK, I've updated the readmes. Hows that?

Alexei-Kornienko:
@TheButlah looks great. Thanks for working on this.

PR #115, opened 2020-12-30
Add WebAssembly Support

TheButlah:
This is a WIP PR that aims to support WASM as a platform.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,12 +29,16 @@ num-traits = "0.2"
 enum-primitive-derive = "0.1"
 dashmap = "3.11"
 crossbeam = "0.7"
-uuid = { version = "0.8", features = ["v4"] }
 regex = "1"
 lazy_static = "1"
 log = "0.4"
 futures_codec = "0.4"
 async-std = { version = "1", features = ["attributes"], optional = true }
+uuid = { version = "0.8", features = ["v4"] }
+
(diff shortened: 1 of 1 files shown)


TheButlah:
The current status of this is that it successfully gets the core library code to compile (but not tests or benchmarks or examples). 

The reason that the other stuff doesn't work is almost entirely down to the fact that the implementation of block_on in async-std doesn't actually block. It also has a different signature from the non-wasm version. This change in signature is what causes `async_std::test` or `async_std::main` to give compilation errors.

See https://github.com/async-rs/async-std/issues/913.

This is merge able in this state, but if we ever want to use `block_on` in the core library (as opposed to just in benchmark or test code for example), we will be blocked by that linked issue.

Alexei-Kornienko:
If I understand correctly currently compilation for webasm is useless cause it doesn't have any working transport available?
I would prefer not merging it until we'll have some proof that it actually works and does something useful.
Could we first implement new transport type for websocket and check it's functionality for linux build?

TheButlah:
Yes, we can do that. I wasn't sure which one to do first, but I agree - this PR doesn't add any value just yet. It was an initial foray to see what the state of things would be and how difficult it would be to get stuff to work. The good news is that it should fully work as soon as async-rs/async-std#913 is fixed (unclear on the timeframe for that).

Next step is to implement the wss:// transport, as you said. This should probably not be merged until that point.

Issue #73, opened 2020-10-04
Connecting sockets to non-existing endpoints does not work


TheButlah:
In ZMQ, you can connect to an endpoint before any socket has actually bound to that endpoint. That lets you avoid all sorts of ordering issues with sockets. Currently, at least the `SubSocket` will fail to connect to non-existent endpoints. We should instead have the sockets compliant with ZMTP:

From https://rfc.zeromq.org/spec/23/ on general socket semantics:
> All sockets SHALL establish connections opportunistically, that is: they connect to an endpoint asynchronously, and if the connection is broken, SHOULD reconnect after a suitable delay.

From https://rfc.zeromq.org/spec/29/ on Subscriber semantics:
>SHALL create a queue when initiating an outgoing connection to a publisher, and SHALL maintain the queue whether or not the connection is established.

Alexei-Kornienko:
This is the tough one.

Should `connect` call block until a connection is finally established? Talking into account zmq specs it seems that it should just prepare all the internal queues and return immediately. However it makes it really hard to return appropriate errors to user in case of any issues. For example user tries to connect to wrong address with no sockets ever listening. How he is supposed to get an error/timeout from the socket? Should there be some kind of backoff between errors? IMHO all of this needs to be carefully investigated before fixing this issue

TheButlah:
I think modeling it after how libzmq does things, but making things idiomatic for Rust and taking care to make error handling idiomatic, is the right approach. I'm unsure of exactly what that would entail and thats easier said than done, but I'll start by enumerating some relevant details I can dig up from libzmq, which may help us figure out what our approach should be.

From http://api.zeromq.org/4-3:zmq-connect
>for most transports and socket types the connection is not performed immediately but as needed by ØMQ. Thus a successful call to zmq_connect() does not mean that the connection was or could actually be established. Because of this, for most transports and socket types the order in which a server socket is bound and a client socket is connected to it does not matter. The ZMQ_PAIR sockets are an exception, as they do not automatically reconnect to endpoints.

From that same link:
>following a zmq_connect(), for socket types except for ZMQ_ROUTER, the socket enters its normal ready state. By contrast, following a zmq_bind() alone, the socket enters a mute state in which the socket blocks or drops messages according to the socket type, as defined in zmq_socket(3). A ZMQ_ROUTER socket enters its normal ready state for a specific peer only when handshaking is complete for that peer, which may take an arbitrary time.

[...]

PR #117, opened 2021-01-07
Implement connect_forever in util. Related to #73

Alexei-Kornienko:
This enables us to connect socket to server that doesn't exist yet.

### DIFF
--- a/src/transport/ipc.rs
+++ b/src/transport/ipc.rs
@@ -13,10 +13,10 @@ use crate::task_handle::TaskHandle;
 use crate::ZmqResult;
 
 use futures::{select, FutureExt};
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
-pub(crate) async fn connect(path: PathBuf) -> ZmqResult<(FramedIo, Endpoint)> {
-    let raw_socket = UnixStream::connect(&path).await?;
+pub(crate) async fn connect(path: &Path) -> ZmqResult<(FramedIo, Endpoint)> {
+    let raw_socket = UnixStream::connect(path).await?;
     let peer_addr = raw_socket.peer_addr()?;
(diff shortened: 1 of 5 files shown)


Alexei-Kornienko:
@TheButlah pls review when you have time. This MR partially fixes #73.

src/transport/ipc.rs
```diff
@@ -15,8 +15,8 @@ use crate::ZmqResult;
 use futures::{select, FutureExt};
 use std::path::{Path, PathBuf};
 
-pub(crate) async fn connect(path: PathBuf) -> ZmqResult<(FramedIo, Endpoint)> {
-    let raw_socket = UnixStream::connect(&path).await?;
+pub(crate) async fn connect(path: &PathBuf) -> ZmqResult<(FramedIo, Endpoint)> {
```
TheButlah, line 18: If we don't need ownership of the `PathBuf`, then this should just be an `&Path`

TheButlah (changes requested):
In order to close #73 , we need connect to return immediately, queuing messages until we run out of memory (or hit the currently unimplemented high water mark). Right now the code instead blocks until the connection is successful. This is an improvement, and we can definitely merge it in this state, but let's not close #73 until we implement HWM functionality and make the connect function non-blocking

src/transport/ipc.rs
Alexei-Kornienko: Thanks for the hint. I'll fix it

Alexei-Kornienko:
I agree. I marked this MR as related cause some additional work is needed to make it fully spec compliant.

src/transport/ipc.rs
Alexei-Kornienko: Done

TheButlah (approved):

PR #120, opened 2021-01-09
Alex/send recv multipart messages

### LINKED ISSUE #119: Allow all socket types to send multipart messages
As far as I can see from the current code REQ/REP and PUB/SUB sockets can't send multipart messages.  That seems to be inconsistent with the spec.  I have made some changes [here](https://github.com/nhuff/zmq.rs/tree/multipart) to support sending multipart messages on all sockets.  The change is fairly intrusive since it changes the ZmqMessage struct to be a wrapper around a VecDeque<Bytes> instead of just a Bytes field.  It also changes all the send/recv methods to take and return ZmqMessage structs.  I figured it might be worth discussion before I opened a really big pull request.

Alexei-Kornienko:
Closes #119

### DIFF
--- a/examples/message_broker.rs
+++ b/examples/message_broker.rs
@@ -20,22 +20,22 @@ async fn main() -> Result<(), Box<dyn Error>> {
         .expect("Failed to bind");
     loop {
         select! {
-            router_mess = frontend.recv_multipart().fuse() => {
+            router_mess = frontend.recv().fuse() => {
                 dbg!(&router_mess);
                 match router_mess {
                     Ok(message) => {
-                        backend.send_multipart(message).await?;
+                        backend.send(message).await?;
                     }
                     Err(_) => {
                         todo!()
                     }
                 }
             },
-            dealer_mess = backend.recv_multipart().fuse() => {
+            dealer_mess = backend.recv().fuse() => {
                 dbg!(&dealer_mess);
                 match dealer_mess {
                     Ok(message) => {
-                        frontend.send_multipart(message).await?;
+                        frontend.send(message).await?;
                     }
                     Err(_) => {
                         todo!()
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -44,22 +44,12 @@ impl GenericSocketBackend {
         loop {
             let next_peer_id = match self.round_robin.pop() {
                 Ok(peer) => peer,
-                Err(_) => match message {
-                    Message::Greeting(_) => panic!("Sending greeting is not supported"),
-                    Message::Command(_) => panic!("Sending commands is not supported"),
-                    Message::Message(m) => {
-                        return Err(ZmqError::ReturnToSender {
-                            reason: "Not connected to peers. Unable to send messages",
-                            message: m,
-                        })
-                    }
-                    Message::Multipart(m) => {
-                        return Err(ZmqError::ReturnToSenderMultipart {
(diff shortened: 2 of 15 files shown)


Alexei-Kornienko:
In subsequent commits we may remove ZmqMessage completely cause now it's only a thin wrapper over bytes and we might not need it at all

TheButlah:
It looks like this exposes Message::Command and the other internal variants. I don't think we should expose this to the end user as it's an internal representation

TheButlah:
What are your thoughts on using a trait for conversion in the public api?

Alexei-Kornienko:
@TheButlah I thought that this might be useful for some future complex scenario when some use would also be able to send commands to somehow manipulate socket options for example. Do we have an RFC in zeromq somewhere that actually explains how commands work? Cause I was not able to find it quickly.

TheButlah:
i thought commands were simply part of the ZMTP protocol but not actually part of the user level API?

Alexei-Kornienko:
>What are your thoughts on using a trait for conversion in the public api?

you mean something like `IntoZmqMessage`? IMHO it makes API a little less obvious cause you need to refer docs to check what types actually support this conversion. I'm not sure how often such approach is used in other crates. Is there some kind of a best practice guide for this?

>i thought commands were simply part of the ZMTP protocol but not actually part of the user level API?

That's the problem I don't actually know what commands are supposed to do. Client socket needs to be able to set up and identity for itself but from what I know it's done during handshake and should be done via socket options. But there may be other stuff that is sent via commands

Alexei-Kornienko:
Looked over all RFCs again. It seems that commands are only mentioned as part of different security mechanisms so there should be visible to user. In such case I agree that exposing Message enum as a parameter is a good option.

Alexei-Kornienko:
>What are your thoughts on using a trait for conversion in the public api?

@TheButlah another thing to take into account is that we have a ReturnToSender error. We should not drop user messages that we cannot send and we should return them via this error. If we use trait as a parameter what type should we actually return? cause it might not be a good idea to return plain Bytes to user (especially if we prepended them with some envelope)

TheButlah:
>Is there some kind of a best practice guide for this?

We don't want to go crazy with traits as arguments because it adds a small amount of mental complexity, but used tastefully it should be fine. In fact, you're already using it prolificly in tokio via the [bind](https://docs.rs/tokio/1.0.1/tokio/net/struct.TcpListener.html#method.bind) function. Here's some additional common examples from popular crates that use this approach:
- https://docs.rs/reqwest/0.11.0/reqwest/fn.get.html
- https://docs.rs/zmq/0.9.2/zmq/struct.Socket.html#method.send and https://docs.rs/zmq/0.9.2/zmq/struct.Socket.html#method.send_multipart
- Anything that uses IntoIterator
- Plenty examples in std lib: 
- Many more, didn't want to list them all.

TheButlah:
>If we use trait as a parameter what type should we actually return? cause it might not be a good idea to return plain Bytes to user (especially if we prepended them with some envelope)

This is the most perplexing problem. We know that internally we have the `Message` enum, which we shouldn't expose to the user. So since we need to return their sent message(s) in the event of an error, there has to be some way of returning that.

I know there are a lot of draft PRs floating around, but I'm going to submit one and let me know what you think.

Alexei-Kornienko:
Obsolete. #122 merged instead of this

PR #122, opened 2021-01-10
Make all socket types support multipart messages.

### LINKED ISSUE #119: Allow all socket types to send multipart messages

nhuff:
The specification says all messages can contain multiple frames.  This
commit gets rid of the distinction between multipart and non multipart
messages.  It does this by having all user facing methods take and
return ZmqMessage structures.  I also modified that structure to be a
wrapper around a VecDeque of Bytes.

Fixes #119

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -1,5 +1,5 @@
 use bytes::Bytes;
-use std::collections::vec_deque::{VecDeque,Iter};
+use std::collections::vec_deque::{Iter, VecDeque};
 use std::convert::{From, TryFrom};
 use std::fmt;
 
@@ -8,7 +8,7 @@ pub struct ZmqEmptyMessageError;
 
 impl fmt::Display for ZmqEmptyMessageError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-	write!(f,"Unable to construct an empty ZmqMessage")
+        write!(f, "Unable to construct an empty ZmqMessage")
     }
 }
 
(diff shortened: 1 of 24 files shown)


src/pub.rs
```diff
@@ -32,11 +32,11 @@ pub(crate) struct PubSocketBackend {
 
 impl PubSocketBackend {
     fn message_received(&self, peer_id: &PeerIdentity, message: Message) {
-        let message = match message {
+        let mut message = match message {
             Message::Message(m) => m,
             _ => return,
         };
-        let data: Vec<u8> = message.into();
+        let data: Vec<u8> = message.pop_front().unwrap().to_vec();
```
Alexei-Kornienko, line 39: One of the drawbacks of this approach is that now user can theoretically send and recv message that contain 0 frames. IMHO this is not good and would cause issues with such unwraps. Before user could send an empty message i.e. 1 frame with message of length 0. But currently nothing is preventing him from constructing and empty ZmqMessage and sending it. IMHO we need to think about how this should be handled cause some sockets depend on certain number of frames available (req/rep/router/dealer)

Alexei-Kornienko (commented):
Looks better than the previous version. However I still have some concerns regarding such approach. some comments inline.

src/pub.rs
nhuff: I think that should probably be dealt with on the sending side as a message with 0 frames isn't valid per the spec. Currently if we tried to send an empty message I don't think anything would go out on the wire and so we would never get a message that was empty.  It could probably mess up the socket state machine in some cases though so we might want to have some kind of check on the send method.

I think there are two things we could do if you passed an empty message into the send function.
1. Do nothing. Essentially make the send a no-op.  This is what pyzmq does and makes some logical sense in that we asked the library to send 0 data to the other side.
2. Return an Err on the send.  I kind of lean towards this option since if you wanted the empty send to be a no-op you can just handle the Err and do nothing. If you accidentally tried to send an empty message you could catch it.

Alexei-Kornienko: >message with 0 frames isn't valid per the spec.

That's what actually my idea is about.. if 0 frames isn't valid you shouldn't be able to create such a message...
Lets leverage type system to prevent it.

```
    pub fn new() -> Self {
	Self { frames: VecDeque::new() }
    }
```
We should simply not allow empty constuctor to the user. So the message could only be created from a certain data that would be a first frame

nhuff: OK I removed the constructors for empty messages.  This doesn't prevent someone from popping all of the frames from the message and getting an empty message.  Not sure if it is worth trying to prevent that or what the correct thing to do in that case is so it doesn't completely solve the issue of send getting an empty message, but it should hopefully make it harder to do by accident.

Alexei-Kornienko: > This doesn't prevent someone from popping all of the frames from the message and getting an empty message.

Good point.. so there is still a way to misuse the API and we should either prevent it or add some more checks to handle this struct.
@TheButlah what do you think about this MR overall?

Alexei-Kornienko:
I've read this MR one more time and I have some new thoughts I would like to share:
* What I would like to see is “make illegal states unrepresentable” I think it's the way how Rust libraries should be written.
* IMHO we can do the following:
1) make pop_front, pop_back pub(crate). By doing this user can only create a message from certain data (e.g. cannot create message with zero frames). And user cannot pop frames out of the message to make it zero size. So in result the message will always be valid cause it would have at least one frame.
2) User can still easily append new parts to message cause he would have push_front, push_back methods available
3) If user needs to shrink the message we should provide method to_deque(self) this method will consume message and return a vecdeque from parts. User can manipulate it as needed and when he want's to send it he would need to convert it back to Message. Which we can do by implementing TryFrom<VecDeque> there we can do all the checks and convert it to valid Message or return an error.

nhuff:
If you want to take a look I think the latest commits should address the raised issues.

src/message.rs
```diff
@@ -1,59 +1,113 @@
...
+
+    pub fn is_empty(&self) -> bool {
+        self.frames.is_empty()
+    }
+
+    pub fn get(&self, index: usize) -> Option<&Bytes> {
+        self.frames.get(index)
+    }
+
+    pub fn to_vec(&self) -> Vec<Bytes> {
```
Alexei-Kornienko, line 49: IMHO this method should take self by value. This will allow to avoid cloning data. Message can be consumed instead

Alexei-Kornienko (commented):
Most of it seems reasonable to me. I'll take a little more time on a weekend to go through it in details.
However I would like to wait for @TheButlah 's review before merging cause I might miss something.

TheButlah:
Sorry this week has been really busy so I've been mostly absent from all of this. @Alexei-Kornienko I will review probably tomorrow or this weekend, but feel free to merge before then if you feel its appropriate

src/message.rs
nhuff: I think I made the change you suggested.  It appears clippy doesn't like it though.  It wants to_xxx to take self by reference.

Alexei-Kornienko: I guess they should be named `into_vec` to make clippy happy.

Alexei-Kornienko: Overall LGTM to me through. I think we can merge this tomorrow

nhuff: I will change the function names.  Do you want me to rebase and squash the commits before you merge?

Alexei-Kornienko: Would be great if you cood rebase to make commit history nicer.
Thank you for your contribution and patience.

nhuff: I will work on getting it cleaned up tomorrow.

Alexei-Kornienko:
Consider it's good enough to be merged. Some minor issues will be fixed in subsequent requests.

PR #135, opened 2021-03-09
Disable Nagle on all sockets

CorinJG:
I think these additional lines in `transport/tcp.rs` are all that's required.


CorinJG:
#135

CorinJG:
#135

PR #136, opened 2021-03-13
Fixed Rep socket panic when message comes from Dealer.

### LINKED ISSUE #132: Rep socket panics when message comes from Dealer
I get panic on message arrival when Rep socket is connected to Dealer socket. Sending message directly from Req to Rep works fine.

```
thread 'tokio-runtime-worker' panicked at 'assertion failed: m.pop_front().unwrap().is_empty()', /home/karmo/.cargo/git/checkouts/zmq.rs-12939e22dc7edef4/511240e/src/rep.rs:144:29
```

This is debugged message before panic:
```
ZmqMessage {
    frames: [
        b"\xf9\x89\xd5\x97x\x9c@\x9c\x9f\xa5\xd4\xb6\xc7\xcb\xd0\xb6",
        b"",
        b"Hello",
    ],
}
```

Sending message like this: `req_socket.send("Hello".into()).await?;`

Setup: Req -> Router <- proxy() -> Dealer <- Rep

I managed to fix it locally. Planning to open a PR soon.

karmux:
Fixes #132

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -53,6 +53,32 @@ impl ZmqMessage {
     pub fn into_vecdeque(self) -> VecDeque<Bytes> {
         self.frames
     }
+
+    pub fn prepend(&mut self, message: &ZmqMessage) {
+        for frame in message.iter().rev() {
+            self.push_front(frame.clone());
+        }
+    }
+
+    pub fn split_at(&self, mid: usize) -> (ZmqMessage, ZmqMessage) {
+        let mut frames_left = VecDeque::with_capacity(mid);
+        let mut frames_right = VecDeque::with_capacity(self.len() - mid);
+        for (index, frame) in self.iter().enumerate() {
+            if index < mid {
+                frames_left.push_back(frame.clone());
+            } else {
+                frames_right.push_back(frame.clone());
+            }
+        }
+        (
+            ZmqMessage {
+                frames: frames_left,
+            },
+            ZmqMessage {
+                frames: frames_right,
+            },
+        )
+    }
 }
 
 impl TryFrom<Vec<Bytes>> for ZmqMessage {
--- a/src/rep.rs
+++ b/src/rep.rs
@@ -12,7 +12,6 @@ use futures::SinkExt;
 use futures::StreamExt;
 use parking_lot::Mutex;
 use std::collections::HashMap;
-use std::convert::TryInto;
 use std::sync::Arc;
 
 struct RepPeer {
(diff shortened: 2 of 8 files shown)


src/rep.rs
```diff
@@ -140,6 +146,10 @@ impl SocketRecv for RepSocket {
                     match message {
                         Message::Message(mut m) => {
                             assert!(m.len() > 1);
+                            if m.len() > 2 {
+                                self.connection_id =
```
Alexei-Kornienko, line 150: Taken from ZMQ spec (https://rfc.zeromq.org/spec/28/)
* An address envelope consisting of zero or more frames, each containing one identity.
For processing incoming messages:
* SHALL remove and store the address envelope, including the delimiter.

So potentially it can be more than one identity in a message. IMHO a better approach would be to store ZmqMessage in the socket - it would contain all frames up to delimiter and later we could just attach reply to it and send it down the wire

Alexei-Kornienko (commented):
Thanks for your MR. I've added some inline comments cause it seems to be a partial fix and not fully compliant to spec. IMHO we should modify MR a little to make it fully compliant.

Alexei-Kornienko:
It might be useful to add some kind of `split_to` method to ZmqMessage to make it easy to split frames apart. As well as `extend` to combine 2 messages into 1

src/message.rs
```diff
@@ -53,6 +53,32 @@ impl ZmqMessage {
...
         self.frames
     }
+
+    pub fn prepend(&mut self, message: &ZmqMessage) {
+        for frame in message.iter().rev() {
+            self.push_front(frame.clone());
+        }
+    }
+
+    pub fn split_at(&self, mid: usize) -> (ZmqMessage, ZmqMessage) {
```
Alexei-Kornienko, line 63: You could use split_off - https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.split_off

src/rep.rs
```diff
@@ -114,6 +116,9 @@ impl SocketSend for RepSocket {
             Some(peer_id) => {
                 if let Some(mut peer) = self.backend.peers.get_mut(&peer_id) {
                     message.push_front(Bytes::from(""));
```
Alexei-Kornienko, line 118: Delimeter should be included in envelope

Alexei-Kornienko (commented):
Just a minor coments otherwise LGTM

Alexei-Kornienko:
Clippy sometimes becomes annoying.. 
LGTM^ minor issues might be fixed in subsequent requests

PR #142, opened 2021-09-24
length of topics are different

youngday:
Error will occurred if topic have different length ,so compare length first.

### DIFF
--- a/src/pub.rs
+++ b/src/pub.rs
@@ -165,7 +165,7 @@ impl SocketSend for PubSocket {
         let mut dead_peers = Vec::new();
         for mut subscriber in self.backend.subscribers.iter_mut() {
             for sub_filter in &subscriber.subscriptions {
-                if sub_filter.len() == message.get(0).unwrap().len(){
+                if sub_filter.len() <= message.get(0).unwrap().len(){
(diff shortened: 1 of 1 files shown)


src/pub.rs
```diff
@@ -165,6 +165,7 @@ impl SocketSend for PubSocket {
         let mut dead_peers = Vec::new();
         for mut subscriber in self.backend.subscribers.iter_mut() {
             for sub_filter in &subscriber.subscriptions {
+                if sub_filter.len() == message.get(0).unwrap().len(){
```
Alexei-Kornienko, line 168: @youngday thanks for the pull request. I agree that this is an issue however IMHO your fix is not quite compatible to ZMQ spec:
"SHALL perform a binary comparison of the subscription against the start of the first frame of the message."
So the proper fix would be to 
sub_filter.len() <= ...

Alexei-Kornienko:
Sorry for slow reply. I'm quite busy lately...

youngday (commented):
changed  with your advise

Alexei-Kornienko:
Minor formatting issues. Otherwise LGTM. Good enough to merge

Alexei-Kornienko (approved):

Issue #147, opened 2021-11-26
[example] How to implement forward zmq device


dimitar-petrov:
Hello,

the following code snippets show my progress so far.

**Forwarder (Message Broker)**
```rust
use std::error::Error;
use zeromq::prelude::*;

use tokio;


#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    let mut frontend = zeromq::SubSocket::new();
    frontend.bind("tcp://127.0.0.1:30001").await?;

    let mut backend = zeromq::PubSocket::new();
    backend.bind("tcp://127.0.0.1:30002").await?;

    frontend.subscribe("").await?;

    let forward = async move {
	loop {
	    let message = frontend.recv().await.unwrap();
	    println!("{:?}", message);
	    backend.send(message).await.unwrap();
	}
    };

    println!("Forwarder start");
    forward.await;

    Ok(())
}

```    
**Publisher**

```rust
use tokio;
use tokio::time::{sleep, Duration};
use std::error::Error;

use zeromq::*;

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    println!("Start client");
    let mut socket = zeromq::PubSocket::new();
    socket.connect("tcp://127.0.0.1:30001").await.expect("Failed to connect");
    sleep(Duration::from_millis(100)).await;

    let mut msg = zeromq::ZmqMessage::from("test");
    // msg.prepend(&"topic".into());

    println!("Send ...");
    socket.send(msg).await?;

    Ok(())
}
```

I can see the message leaving the publisher and coming on the frontend port of forward device (using tcpdump), but do not receive it on the forward device.

Thanks in advance @Alexei-Kornienko

dimitar-petrov:
Update:

Looks like that issue comes from running as SubSocket as server and binding it on the port and running the PubSocket as client.

This example connects both sockets but no messages are received on the subscriber

Subscriber:
```rust
use zeromq::*;
use std::error::Error;
use tokio;
use tokio::time::{
    sleep, Duration
};
use zeromq::SocketRecv;

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    println!("Starting server...");

    let mut socket = zeromq::SubSocket::new();
    socket.bind("tcp://127.0.0.1:30001").await?;
    socket.subscribe("").await?;

    loop {
	let msg = socket.recv().await.unwrap();
	println!("{:?}", msg);
    }

    Ok(())
}
```

Publisher
[...]

PR #149, opened 2021-12-03
feat: add support for reverse binding between PUB and SUB

dimitar-petrov:
In order to provide functionality required by following example #147 a patch to PUB and SUB sockets should be implemented.

Quote from dev on discord:
```
subscriber socket should keep list of his subscriptions and send it to PUB socket when it connects
```

**Todos**
- [X] add forwarder_device.rs example
- [X] add subs HashSet
- [X] convert peer_connected to async fn and update all referencing structs
- [x] sent all subscriptions from the HashSet on peer connect
- [x] refactor
- [x] fix code format and style

Any comments regarding possible solutions are appreciated.
Thanks.

### DIFF
--- a/src/sub.rs
+++ b/src/sub.rs
@@ -5,22 +5,89 @@ use crate::message::*;
 use crate::transport::AcceptStopHandle;
 use crate::util::PeerIdentity;
 use crate::{
-    MultiPeerBackend, Socket, SocketBackend, SocketEvent, SocketOptions, SocketRecv, SocketType,
+    MultiPeerBackend, Socket, SocketBackend, SocketEvent, SocketOptions, SocketRecv, SocketType
 };
 
-use crate::backend::GenericSocketBackend;
+use crate::backend::Peer;
+use dashmap::DashMap;
+use parking_lot::Mutex;
+use crate::fair_queue::QueueInner;
 use crate::fair_queue::FairQueue;
+use crossbeam::queue::SegQueue;
 use async_trait::async_trait;
 use bytes::{BufMut, BytesMut};
 use futures::channel::mpsc;
 use futures::{SinkExt, StreamExt};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
+pub(crate) struct SubSocketBackend {
+    pub(crate) peers: DashMap<PeerIdentity, Peer>,
+    fair_queue_inner: Option<Arc<Mutex<QueueInner<ZmqFramedRead, PeerIdentity>>>>,
+    pub(crate) round_robin: SegQueue<PeerIdentity>,
+    socket_type: SocketType,
+    socket_options: SocketOptions,
+    pub(crate) socket_monitor: Mutex<Option<mpsc::Sender<SocketEvent>>>,
+}
+
+impl SubSocketBackend {
+    pub(crate) fn with_options(
+        fair_queue_inner: Option<Arc<Mutex<QueueInner<ZmqFramedRead, PeerIdentity>>>>,
+        socket_type: SocketType,
+        options: SocketOptions,
+    ) -> Self {
+        Self {
+            peers: DashMap::new(),
+            fair_queue_inner,
+            round_robin: SegQueue::new(),
+            socket_type,
+            socket_options: options,
+            socket_monitor: Mutex::new(None),
+        }
+    }
+}
+
+impl SocketBackend for SubSocketBackend {
+    fn socket_type(&self) -> SocketType {
+        self.socket_type
+    }
+
+    fn socket_options(&self) -> &SocketOptions {
+        &self.socket_options
+    }
+
+    fn shutdown(&self) {
+        self.peers.clear();
+    }
+
+    fn monitor(&self) -> &Mutex<Option<mpsc::Sender<SocketEvent>>> {
+        &self.socket_monitor
+    }
+}
+
+impl MultiPeerBackend for SubSocketBackend {
+    fn peer_connected(self: Arc<Self>, peer_id: &PeerIdentity, io: FramedIo) {
+        let (recv_queue, send_queue) = io.into_parts();
+        self.peers.insert(peer_id.clone(), Peer { send_queue });
+        self.round_robin.push(peer_id.clone());
+        match &self.fair_queue_inner {
+            None => {}
+            Some(inner) => {
+                inner.lock().insert(peer_id.clone(), recv_queue);
+            }
+        };
+    }
+
+    fn peer_disconnected(&self, peer_id: &PeerIdentity) {
+        self.peers.remove(peer_id);
+    }
+}
+
 pub struct SubSocket {
-    backend: Arc<GenericSocketBackend>,
+    backend: Arc<SubSocketBackend>,
     fair_queue: FairQueue<ZmqFramedRead, PeerIdentity>,
     binds: HashMap<Endpoint, AcceptStopHandle>,
+    subs: HashSet<String>,
 }
 
 impl Drop for SubSocket {
@@ -41,10 +108,12 @@ impl SubSocket {
                 .send(Message::Message(message.clone()))
                 .await?;
         }
+	self.subs.insert(subscription.to_string());
         Ok(())
     }
 
     pub async fn unsubscribe(&mut self, subscription: &str) -> ZmqResult<()> {
+	self.subs.insert(subscription.to_string());
         let mut buf = BytesMut::with_capacity(subscription.len() + 1);
         buf.put_u8(0);
         buf.extend_from_slice(subscription.as_bytes());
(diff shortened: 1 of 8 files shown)


Alexei-Kornienko:
Basically we need to implement 2 things:
1) https://github.com/zeromq/zmq.rs/blob/master/src/sub.rs#L23 Here we should add a HashSet<String> to keep subscriptions and add/remove from it with each `subscribe` `unsubscribe` call (https://github.com/zeromq/zmq.rs/blob/master/src/sub.rs#L33)

2) this part is more complicated. I Have a struct called `backend` to encapsulate most of the socket logic related to network. Backend provides call on_peer_connected which is called when a new connection is established and handshake completes (https://github.com/zeromq/zmq.rs/blob/master/src/backend.rs#L101). SUB now uses `GenericSocketBackend` We need to replace it with a custom backend for SUB socket that will use HashSet created on step 1 to send current subscriptions to PUB socket when it first connects. This can be done by either copy/pasting GenericSocketBackend impl to sub socket and customizing it or by using composition of structs

dimitar-petrov:
Couple of questions:

1. How to access `SubSocket.subs` (HashSet) from `MultiPeerBackend.peer_connected`, since it is Composite?
2. What about putting the `HashSet` as property of `SubSocketBackend` struct and acessing it as `self.backend.subs` from `subscribe` and `unsubscribe`?

Alexei-Kornienko:
2. Yeah you might be right. I guess in such case it might be easier to put it in backend itself. would need to use a mutex for it but it should not be a problem cause it's not on a fast path of the code.

Another issue I see now is that peer_connected is not async. It means that you want be able to send messages to pub socket and await them. I see 2 possible options:
* we have try_send method that relies on eventual flush (However It may not be 100% reliable)
* make peer_connected async - should be easy except that all other places would also be updated

src/sub.rs
```diff
@@ -41,10 +108,12 @@ impl SubSocket {
                 .send(Message::Message(message.clone()))
                 .await?;
         }
+	self.subs.insert(subscription.to_string());
         Ok(())
     }
 
     pub async fn unsubscribe(&mut self, subscription: &str) -> ZmqResult<()> {
+	self.subs.insert(subscription.to_string());
```
Alexei-Kornienko, line 116: I guess should be remove here

dimitar-petrov: Sure

src/sub.rs
```diff
@@ -31,6 +116,7 @@ impl Drop for SubSocket {
 
 impl SubSocket {
     pub async fn subscribe(&mut self, subscription: &str) -> ZmqResult<()> {
```
Alexei-Kornienko, line 118: I guess subscribe/unsubscribe impl can be moved to backend to avoid copy/pasting code around

Alexei-Kornienko:
Mostly LGTM. I would suggest minor refactoring + maybe run a fmt/clippy to make sure that code looks pretty

src/sub.rs
```diff
@@ -8,17 +8,102 @@ use crate::{
...
+        self.peers.insert(peer_id.clone(), Peer { send_queue });
+        self.round_robin.push(peer_id.clone());
+        match &self.fair_queue_inner {
+            None => {}
+            Some(inner) => {
+                inner.lock().insert(peer_id.clone(), recv_queue);
+            }
+        };
+
+	let subs = (|| { (*self.subs.lock()).clone() })();
```
dimitar-petrov, line 84: This feels ugly. Wha is an idiomatic way to get the data and unlock the mutex.

Alexei-Kornienko: You can just write lock().clone() With modern rust borrow cherker mutex will be released on the same line

src/sub.rs
dimitar-petrov: Could you elaborate a little bit on that.

src/sub.rs
Alexei-Kornienko: Even a better way of doing this would be to create a Vec of ZmqMessages with the mutex locked (this will avoid extra copy). So IMHO a better solution would be to:
1) take lock
2) iterate subs and create ZmqMessage for each
3) release lock
4) send messages

src/sub.rs
Alexei-Kornienko: I don't really like that we have 2 places where we create a subscribe message. So I would look for a way to move this code to some common function.

src/sub.rs
```diff
@@ -8,17 +8,102 @@ use crate::{
...
+
+	for sub in subs.iter() {
+	    let mut buf = BytesMut::with_capacity(sub.len() + 1);
+	    buf.put_u8(1);
+	    buf.extend_from_slice(sub.as_bytes());
+
+	    let message: ZmqMessage = ZmqMessage::from(buf.freeze());
+	    let mut peer = self.peers.get_mut(&peer_id).unwrap();
```
Alexei-Kornienko, line 92: you can use io: FramedIo parameter to send messages directly to this peer. There is no need to do a lookup for it

src/sub.rs
dimitar-petrov: I went with the following approach but cannot borrow due to this: `E0596 cannot borrow data in an 'Arc' as mutable`
```rust
pub enum SubBackendMessageType {
    UNSUBSCRIBE = 0,
    SUBSCRIBE = 1
}

impl SubSocket {
    pub async fn subscribe(&mut self, subscription: &str) -> ZmqResult<()> {
	self.backend.handle_subscription(
	    subscription, SubBackendMessageType::SUBSCRIBE).await
    }

    pub async fn unsubscribe(&mut self, subscription: &str) -> ZmqResult<()> {
	self.backend.handle_subscription(
	    subscription, SubBackendMessageType::UNSUBSCRIBE).await
    }
}
```
and defined `handle_subscription` on the backend
```rust
    pub async fn handle_subscription(
	&mut self, subscription: &str, msg_type: SubBackendMessageType) -> ZmqResult<()> {
        let mut buf = BytesMut::with_capacity(subscription.len() + 1);
        buf.put_u8(msg_type as u8);
        buf.extend_from_slice(subscription.as_bytes());

        let message: ZmqMessage = ZmqMessage::from(buf.freeze());
        for mut peer in self.peers.iter_mut() {
            peer.send_queue
                .send(Message::Message(message.clone()))
                .await?;
        }
        self.subs.lock().insert(subscription.to_string());

	Ok(())
	
    }
```
How would you define a common function in the backend and call it from the `SubSocket`

Thanks

Alexei-Kornienko: 1) Backend is shared between threads so it cannot have methods that take `&mut self`.
2) I would simple create an assosiated function that would create a subscribe message in it.
```
fn subscribe(subscription: &str) -> ZmqMessage {
        let mut buf = BytesMut::with_capacity(subscription.len() + 1);
        buf.put_u8(msg_type as u8);
        buf.extend_from_slice(subscription.as_bytes());

        buf.freeze().into()
}
```
And use it where we need to create a message.

Alexei-Kornienko:
Overall LGTM. Could you please fix formatting issues reported by CI so we can get it merged?

dimitar-petrov:
Sure I will take care of formatting and push:

I have two questions in the mean time:
1. Why this snippet is not working. For some reason it is not unlocking the mutex and compiler complains due to awaiting latter in the function.
```rust
let subs = self.subs.lock();
  let subs_msgs: Vec<ZmqMessage> = subs.iter().map(
      |x| SubSocketBackend::create_subs_message(
	  &x, SubBackendMsgType::SUBSCRIBE)).collect();
  drop(subs);
```
2. And I noticed that if I disconnect and reconnect the sub server process, a client with the following code does not detect anytihg and is trying to push messages.
```rust
#[tokio::main]
    println!("Start client");
    let mut socket = zeromq::PubSocket::new();
    socket.connect("tcp://127.0.0.1:30001").await.expect("Failed to connect");
    sleep(Duration::from_millis(100)).await;

    let mut msg = zeromq::ZmqMessage::from("test");
    msg.prepend(&"topic".into());

    loop {
	println!("Send ...");
	println!("{:?}", socket.send(msg.clone()).await.unwrap());
	sleep(Duration::from_millis(1000)).await
    }

    Ok(())
}
```
```

**Comment by Alexei-Kornienko** [comment]
1. Seems like compiler issue. Following code does work:
```
        let subs_msgs = {
            let subs = self.subs.lock();
            let subs_msgs: Vec<ZmqMessage> = subs.iter().map(
                  |x| SubSocketBackend::create_subs_message(
                      &x, SubBackendMsgType::SUBSCRIBE)).collect();
            drop(subs);
            subs_msgs
        };

```
So in this case I would say that compiler rejects correct code.

2. That's another open issue I have in the library (#143 ). Code currently doesn't handle reconnect on it's own as it should according to ZMQ spec. However IMHO it should be fixed via separate MR.

**Comment by dimitar-petrov** [comment]
@Alexei-Kornienko, let me know if you are fine with the function/variable naming conventions I used.

File: src/sub.rs [hunk 5]
```diff
@@ -8,17 +8,114 @@ use crate::{
...
+        let subs_msgs: Vec<ZmqMessage> = self
+            .subs
+            .lock()
+            .iter()
+            .map(|x| SubSocketBackend::create_subs_message(x, SubBackendMsgType::SUBSCRIBE))
+            .collect();
+
+        for message in subs_msgs.iter() {
+            send_queue
+                .send(Message::Message(message.clone()))
```
Alexei-Kornienko, line 97: just a minor comment. There is no need to clone message here. We are sending each message only once. You can use into_iter and send message by value

Alexei-Kornienko:
LGTM.

PR #150, opened 2021-12-18
Handle errors in Pull and Router

njeisecke:
ConnectionReset is a common error that occurs here.

Invoking peer_disconnected regardless of the error might be a little coarse but that should be better than raising a panic.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -114,5 +114,11 @@ impl MultiPeerBackend for GenericSocketBackend {
 
     fn peer_disconnected(&self, peer_id: &PeerIdentity) {
         self.peers.remove(peer_id);
+        match &self.fair_queue_inner {
+            None => {}
+            Some(inner) => {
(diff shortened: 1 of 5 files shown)


njeisecke:
Seems this is not enough.

Could it be that GenericSocketBackend::peer_disconnect must also remove the peer from fair_queue?

njeisecke:
@Alexei-Kornienko, could you please confirm (or deny) this being the proper way to handle errors? Thanks.

Alexei-Kornienko:
Mostly LGTM. I guess we can merge this. Could you please remove dbg! from the code? I think that we should use some other way to log error instead of just writing it to stderr (maybe use `tracing` crate?).

njeisecke:
Thanks for reviewing. I think using the `tracing` crate would be a good idea so that the application can decide how to log this. For now I've removed the `dbg!`.

PR #156, opened 2022-04-19
[WIP] Feat: Support TCP Keep-alive options

### LINKED ISSUE #155: Tcp keep alive setting function
Hi all, I use this crate on receiving live stock quotes from stock exchanges. When the market daily rests for a while and then restarts trading, I could get no data from the zmq connection established before. 
It can be fixed by tcp keep alive settings, but I could not find the related function in this crate. Could you help me with it?

poyea:
We may need to re-think the design of this option on how each of these options should be supported.

- [ ] Enhancement on basic options, e.g. types / defaults
- [ ] TCP Keep-alive options with examples
- [ ] Tests
- [ ] Docs

Related:
* http://api.zeromq.org/4-0:zmq-setsockopt#toc30
* https://github.com/zeromq/libzmq/blob/master/src/options.cpp#L504

Resolves #155.

### DIFF
--- a/examples/socket_client_with_options.rs
+++ b/examples/socket_client_with_options.rs
@@ -2,13 +2,20 @@ mod async_helpers;
 
 use std::convert::TryFrom;
 use std::error::Error;
-use zeromq::util::PeerIdentity;
+use zeromq::util::{PeerIdentity, TcpKeepalive};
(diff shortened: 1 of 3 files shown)


rgbkrk:
I'm closing old PRs. Let me know if you want to bring this back again later.

PR #160, opened 2022-07-14
Remove unnecessary allocations and unsafe

andrewwhitehead:


src/codec/command.rs
```diff
@@ -46,27 +53,29 @@ impl ZmqCommand {
     }
 }
 
-impl TryFrom<BytesMut> for ZmqCommand {
+impl TryFrom<Bytes> for ZmqCommand {
```
andrewwhitehead, line 56: Switched to Bytes for consistency

src/codec/command.rs
```diff
@@ -46,27 +53,29 @@ impl ZmqCommand {
...
     type Error = CodecError;
 
-    fn try_from(mut buf: BytesMut) -> Result<Self, Self::Error> {
+    fn try_from(mut buf: Bytes) -> Result<Self, Self::Error> {
         let command_len = buf.get_u8() as usize;
         // command-name-char = ALPHA according to https://rfc.zeromq.org/spec:23/ZMTP/
-        let command_name =
-            unsafe { String::from_utf8_unchecked(buf.split_to(command_len).to_vec()) };
```
andrewwhitehead, line 56: This is UB given that the peer may not be following the spec.

PR #163, opened 2022-08-06
change  zmq into zmq2

youngday:
now ,zmq are not update ,replaced by zmq2,  zmq2 <- zmq-sys2 <- libzmq-src, now v4.3.4


Alexei-Kornienko:
LGTM

PR #164, opened 2022-08-12
Use Bytes inside PeerIdentity

andrewwhitehead:
This makes cloning of PeerIdentity (which is performed quite often) much cheaper.

Also adds a `from_slice` method and a `FromStr` implementation.


Alexei-Kornienko:
LGTM

PR #170, opened 2022-08-15
Clean up futures-* dependencies

andrewwhitehead:
Avoids using the parent `futures` crate so that dependencies are clearer, and imports a few things from `std::task` that are aliased in `futures`.


poyea (approved):
👍

PR #172, opened 2022-09-06
Handle codec errors in REQ socket recv

nullsauce:
Hey there. 
This PR handles codec errors when reading from a REQ socket and thus makes the code more resilient to abrupt changes in the underlying tcp connection. In my experience, it was relatively easy to make the panic happen during tests.

Not sure what's planned for the ` Some(Ok(_))` arm, so I've left it as `todo()`.

Hope this helps. Please let me know if I should do anything particular with the PR. Cheers!


mmastrac (approved):
This seems reasonable to me. I suspect there may need to be some tests added around underlying TCP socket issues.

PR #178, opened 2023-04-30
Only use required features of regex crate

davidlattimore:
This almost halves the time required to compile regex.


davidlattimore:
On my laptop, regex with default features takes 8.2 seconds to compile. With only "std" and "unicode-perl" this drops to 4.6 seconds.

PR #180, opened 2023-04-30
Automatically derive a Default impl

davidlattimore:
Deriving Default for enums was stabilized in 1.62.0.


davidlattimore:
I'm also happy to send a PR to just suppress the warning if you'd rather not bump the MSRV.

mmastrac (approved):
LGTM!

PR #185, opened 2024-01-15
only build ipc transport on *nix systems

rgbkrk:
Hey all! I'm hoping to get a zeromq backed Tauri app to build on all systems. See https://github.com/nteract/nteract-next/pull/16 for the app and the current PR.

I've added in a check for both the `ipc-transport` feature flag and checking that the platform is a *nix system. Let me know if there's a better approach for this.


mmastrac (approved):
Looks good to me!

PR #187, opened 2024-02-05
feat: Allow overriding block_on_no_clients queue param

bemyak:
In the current implementation, when a SUB provider is restarted, the library just hangs without detecting connection loss. This happens because on disconnect a client is kicked out of the queue, as since for all the socket queues block_on_no_clients is set to true, it hangs indefinitely.

I'm unsure what would be the effect of setting block_on_no_clients to false globally (though it seems reasonable), so in this PR add an ability to override this value is introduced via SocketOptions.

### DIFF
--- a/src/dealer.rs
+++ b/src/dealer.rs
@@ -31,7 +31,7 @@ impl Drop for DealerSocket {
 #[async_trait]
 impl Socket for DealerSocket {
     fn with_options(options: SocketOptions) -> Self {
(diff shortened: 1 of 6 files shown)


rgbkrk:
Hey @bemyak, thanks for the contribution and sorry it's taken so long to look over. I'm going to go ahead and rebase this to make sure it builds against latest.

rgbkrk:
I thought for sure that would trigger CI.

PR #193, opened 2024-05-31
enable async dispatcher macros only for tests

rgbkrk:
Part of this is a code cleanup to gate the tests behind `cfg(test)`. The rest is to enable the async dispatcher macros for the tests (defining `main` and `test` macros for handling) and ensuring they aren't exposed for downstream use.


rgbkrk:
Going to come back to this one to make test support a more general macro.

PR #198, opened 2024-10-05
Added support for futures-task v0.3.31

PaoloPana:
As described [in the changelog](https://github.com/rust-lang/futures-rs/blob/master/CHANGELOG.md#0331---2024-10-05) and [here](https://github.com/rust-lang/futures-rs/pull/2830) with the latest version of futures-task they update the function `waker_ref` making it 'static. For this, zmq needs also to add 'static to some properties.


rgbkrk:
Thank you for correcting this!

PR #217, opened 2025-11-06
update rand v0.9.x

jessekrubin:
I noticed that rand is on v0.8 so here is a PR to update to 9

### DIFF
--- a/examples/stock_server.rs
+++ b/examples/stock_server.rs
@@ -6,7 +6,7 @@ use zeromq::*;
 
 #[async_helpers::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut rng = rand::thread_rng();
+    let mut rng = rand::rng();
     let stocks: Vec<&str> = vec!["AAA", "ABB", "BBB"];
     println!("Starting server");
     let mut socket = zeromq::PubSocket::new();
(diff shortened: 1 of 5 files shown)


examples/stock_server.rs
```diff
@@ -6,7 +6,7 @@ use zeromq::*;
 
 #[async_helpers::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut rng = rand::thread_rng();
+    let mut rng = rand::rng();
```
rgbkrk: Since I wasn't familiar with what was deprecated I'm going to go through each of these and mark with a 👍

examples/stock_server.rs
```diff
@@ -15,7 +15,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Start sending loop");
     loop {
         for stock in &stocks {
-            let price: u32 = rng.gen_range(1..100);
+            let price: u32 = rng.random_range(1..100);
```
rgbkrk: 👍

examples/task_ventilator.rs
```diff
@@ -26,10 +26,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
     sink.send("0".into()).await?;
 
-    let mut rnd = rand::thread_rng();
+    let mut rnd = rand::rng();
     let mut total_msec = 0;
     for _ in 0..100 {
-        let workload = rnd.gen_range(1..100);
+        let workload = rnd.random_range(1..100);
         sender.send(workload.to_string().into()).await?;
         total_msec += workload;
```
rgbkrk: 👍

PR #220, opened 2025-12-18
Fix fair_queue tests to match disconnect behavior

rgbkrk:
PR #207 changed FairQueue to return None immediately when any stream ends (to properly handle disconnects). The tests from PR #213 were written with the old behavior in mind where the queue would continue until all streams finished.

Updated tests to expect None when the first stream ends rather than waiting for all streams to complete or return Pending.

cc @jbbjarnason

### DIFF
--- a/src/fair_queue.rs
+++ b/src/fair_queue.rs
@@ -333,12 +333,11 @@ mod test {
             other => panic!("Expected fast stream second, got: {:#?}", other),
         }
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Ok so this surfaced an issue on a REP socket with 100 concurrent clients. The first client to finish would cause the server to stop receiving from the remaining 99 clients.

poyea (approved):

Issue #233, opened 2026-03-09
Investigating memory leak on aarch64 target


NoMore201:
I'm currently writing a zeromq application that is compiled to x86_64, armv7, aarch64. On x86_64 and armv7 everything seems to be working, and I can reach 24-48 hours of uptime with no issue. On aarch64 target (running a custom Yocto/Openembedded linux distribution), I'm experiencing huge and sudden memory leaks after 30min/1h of uptime:

<img width="1403" height="873" alt="Image" src="https://github.com/user-attachments/assets/b92d6f5f-32b5-4749-8a95-d7d4e8583b7b" />

**Fig 1**: CPU and memory usage **monitoring** over time

Memory usage suddenly start growing very quick until the OOM daemon kils the application when it reaches ~50% RAM usage. Running the application with heaptrack, I see that most of the memory allocation originates from zeromq `begin_accept` function.

<img width="1392" height="619" alt="Image" src="https://github.com/user-attachments/assets/fa121cfb-5f99-41b9-8f45-d66c3f6b7c56" />

**Fig 2**: Bottom up view of heaptrack report

Added a few debug println to the `begin_accept` function:

https://github.com/NoMore201/zmq.rs/blob/28c1303fed03de13fbc8d23bb3570148c8b6cfa5/src/transport/tcp.rs#L56-L61

and I tried to reproduce the issue again. It seems that the memory leak is caused by the application reaching limits for used file descriptors:

```
Mar 09 14:22:58 host launch_app.sh[1541427]: Error accepting connection: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
Mar 09 14:22:58 host launch_app.sh[1541427]: accept called
Mar 09 14:22:58 host launch_app.sh[1541427]: Error accepting connection: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
Mar 09 14:22:58 host launch_app.sh[1541427]: accept called
Mar 09 14:22:58 host launch_app.sh[1541427]: Error accepting connection: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
Mar 09 14:22:58 host launch_app.sh[1541427]: accept called
Mar 09 14:22:58 host launch_app.sh[1541427]: Error accepting connection: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
Mar 09 14:22:58 host launch_app.sh[1541427]: accept called
[...]

PR #234, opened 2026-03-10
Added on Disconnect callback on all sockets

Alexei-Kornienko:
Fixes issue #233.

### DIFF
--- a/src/dealer.rs
+++ b/src/dealer.rs
@@ -31,13 +31,22 @@ impl Drop for DealerSocket {
 #[async_trait]
 impl Socket for DealerSocket {
     fn with_options(options: SocketOptions) -> Self {
-        let fair_queue = FairQueue::new(true);
+        let mut fair_queue = FairQueue::new(true);
+        let backend = Arc::new(GenericSocketBackend::with_options(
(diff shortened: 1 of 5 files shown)


poyea (approved):

src/dealer.rs
```diff
@@ -31,13 +31,22 @@ impl Drop for DealerSocket {
 #[async_trait]
 impl Socket for DealerSocket {
     fn with_options(options: SocketOptions) -> Self {
-        let fair_queue = FairQueue::new(true);
+        let mut fair_queue = FairQueue::new(true);
+        let backend = Arc::new(GenericSocketBackend::with_options(
```
poyea, line 35: Minor: there's some code duplication but not a huge concern

Alexei-Kornienko: Yeah I don't like it as well. I'm trying to do a full review of the code structure to look for a better way of doing stuff. But currently I don't have an option that would be significantly better

clouedoc:
I am a bit late to the party, but I can confirm that this PR fixed my issue in production. Huge thanks, @Alexei-Kornienko and @poyea  !

PR #235, opened 2026-04-01
impl `Clone` for `RouterSendHalf`

BugenZhao:
Expose `Clone` for `RouterSendHalf` so callers can share the send half across tasks without adding an extra mutex layer.

Also add router split tests covering:
- `RouterSendHalf: Clone`
- concurrent sends from two cloned `RouterSendHalf`s to the same dealer


Alexei-Kornienko:
LGTM

poyea (approved):

PR #236, opened 2026-04-09
feat: support `XSubSocket`

BugenZhao:
This adds a public `XSubSocket` implementation to match the existing `SocketType::XSUB` support.

The implementation supports bind/connect, recv/send, subscribe/unsubscribe, and reconnect with subscription replay. As part of this, the shared subscription and reconnect logic used by `SUB` and `XSUB` is moved into a common backend while keeping `SubSocket` behavior unchanged. Subscription state is stored as raw bytes so non-UTF8 subscriptions are preserved correctly across unsubscribe and reconnect replay.

The change also adds coverage for basic `XSUB <-> XPUB` behavior, arbitrary `XSUB -> XPUB` sends, fanout, reconnect, proxying, and minimal cross-implementation tests against `libzmq`.

This should address https://github.com/zeromq/zmq.rs/issues/40.

### DIFF
--- a/src/sub_backend.rs
+++ b/src/sub_backend.rs
@@ -205,15 +205,44 @@ impl SubSocketBackend {
 
     /// Reliably send a message to every connected peer using async send.
     async fn broadcast_control_message(&self, message: ZmqMessage) -> ZmqResult<()> {
+        let mut dead_peers = Vec::new();
+        let mut first_error = None;
         let mut iter = self.peers.begin_async().await;
         while let Some(mut peer) = iter {
-            peer.send_queue
+            let peer_id = peer.key().clone();
+            let result = peer
+                .send_queue
                 .as_mut()
                 .send(Message::Message(message.clone()))
-                .await?;
+                .await;
+            match result {
+                Ok(()) => {}
+                Err(e)
+                    if matches!(
+                        &e,
+                        CodecError::Io(io_error) if io_error.kind() == ErrorKind::BrokenPipe
+                    ) =>
+                {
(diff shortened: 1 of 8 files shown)


Alexei-Kornienko:
Thanks for the PR. I will take a closer look on this on the weekend

src/sub_backend.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    /// Reliably send a message to every connected peer using async send.
+    async fn broadcast_control_message(&self, message: ZmqMessage) -> ZmqResult<()> {
+        let mut iter = self.peers.begin_async().await;
+        while let Some(mut peer) = iter {
+            peer.send_queue
+                .as_mut()
+                .send(Message::Message(message.clone()))
+                .await?;
```
Alexei-Kornienko: This `?` is an issue. Failing to send to a single peer will break the loop and cause other healthy peers to miss the message. Proper error handling is needed in this case

Alexei-Kornienko (changes requested):
One issue regarding error handling. Otherwise LGTM

src/sub_backend.rs
BugenZhao: Thanks for pointing out. Fixed in latest commit

Alexei-Kornienko (approved):
LGTM

poyea (approved):

rgbkrk:
Oh my gosh, thank you @Alexei-Kornienko and @poyea for stewarding maintenance on zmq.rs.

Issue #238, opened 2026-04-19
0.5.0: Advertised MSRV seems wrong


birkenfeld:
zeromq 0.5 advertises MSRV 1.62, but requires scc >=3.4.4, which in turn requires sdd>=4.4, which requires Rust 1.85.

poyea:
Confirmed with:
```
git checkout v0.5.0
cargo +stable tree -i saa
cargo +stable metadata --format-version 1 | jq -r '.packages[] | select(.name == "scc" or .name == "sdd" or .name == "saa" or .name == "zeromq") | "\(.name) \(.version) rust_version=\(.rust_version // "N/A") edition=\(.edition)"'
```
which gives
```
saa 5.5.1 rust_version=1.85.0 edition=2024
scc 3.7.0 rust_version=1.85.0 edition=2024
sdd 4.8.6 rust_version=1.85.0 edition=2024
zeromq 0.5.0 rust_version=1.62.0 edition=2021
```

PR #239, opened 2026-04-19
Bump MSRV to 1.85 and add checks to CI

poyea:
This should remediate #238. I add checks of MSRV to CI for future commits and update`Cargo.toml` to fix the advertisement. With this, we can also extend the check to build zmq.rs against nightly.


Alexei-Kornienko (approved):
LGTM

Issue #240, opened 2026-04-26
Heads-up: rustzmq2, an async-native fork of zmq.rs


PoHsuanLai:
Hi all — wanted to flag a fork of zmq.rs I've been working on, in case any of it is interesting upstream. [rustzmq2](https://github.com/PoHsuanLai/rustzmq2) has grown enough in scope that publishing it as its own crate felt cleaner than landing everything as PRs.

**What's different:**

- Reworked I/O engine: per-peer tasks, vectored writes, slab-based peer registry
- Full ZMTP security stack: PLAIN ([RFC 24](https://rfc.zeromq.org/spec/24/)), CURVE ([RFC 25](https://rfc.zeromq.org/spec/25/)), ZAP ([RFC 27](https://rfc.zeromq.org/spec/27/))
- `smol` runtime support (alongside `tokio`)
- New socket types: PAIR, SCATTER, GATHER, CHANNEL
- Roughly on par with libzmq in our benches ([live comparison](https://pohsuanlai.github.io/rustzmq2/comparison.html), [methodology](https://github.com/PoHsuanLai/rustzmq2/blob/master/docs/BENCHMARKS.md))

**Why a fork:** the changes were large and opinionated enough that I wasn't sure they'd match upstream's direction. Happy to carve individual pieces back as PRs if useful

Original authors credited in [CREDITS.md](https://github.com/PoHsuanLai/rustzmq2/blob/master/CREDITS.md). Upstream code retains MIT; fork additions are MIT OR Apache-2.0.

Alexei-Kornienko:
Just out of curiosity: have you compared performance to this library as well?

Alexei-Kornienko:
A quick scan over your code shows that you delete FairQueue primitive and I could find a place where you actually balance recv part (I checked the Rep socket). So it seems to me that your impl. doesn't comply to ZMQ spec in terms of fair recv and a single producer can spam over other peers.

Overall my thoughts are:
1) I always welcome new contributions and possibly new maintainers (if it's going to be approved by other maintainers as well)
2) This project is unfortunately not really high in my priorities list so I cannot spend a lot of time working on it.
3) for your fork diff --stat shows 183 files changed, 24128 insertions(+), 4534 deletions(-). There is no way I would be able to review all this code in any near future.
[...]

PR #246, opened 2026-05-02
Ported benchmarks from libzmq2

Alexei-Kornienko:
I've ported set of benchmarks from rustzmq2 see #240 . This may give us a better baseline for comparison of performance both to libzmq implementation and other rust variants


poyea (approved):
lgtm

benches/bench_runtime.rs
```diff
@@ -0,0 +1,51 @@
...
+            async_std::task::block_on(future)
+        }
+    }
+}
+
+impl Default for BenchRuntime {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
rgbkrk, line 51: We could also handle the async dispatcher runtime in here but I don't think we have to block on that as we'll get good benchmarks out of the two main ones.

PR #249, opened 2026-05-05
Fix fair queue spin when pending streams wake themselves

### LINKED ISSUE #248: PullSocket::recv stalls when wrapped in tokio::time::timeout (regression in 0.5)
## Summary

Wrapping `PullSocket::recv()` in `tokio::time::timeout(_, recv())` per call
causes the receiver to stall after a few thousand messages, even when the
timeout never fires. Removing the timeout wrapper restores normal throughput
(1M+ msg/s for 128 B messages on loopback).

## Bisect

| version | per-call `timeout(_, recv())` | plain `recv()` |
|---------|-------------------------------|----------------|
| 0.4.1   | works (~880k msg/s)           | works (~1M)    |
| 0.5.0   | stalls within ~10k msgs       | works (~1M)    |
| 0.6.0   | stalls within ~10k msgs       | works (~1M)    |

The regression matches the scc + fair_queue rework in 0.5 (#207, #220).

## Repro

`Cargo.toml`:
```toml
[dependencies]
zeromq = "0.6"
tokio = { version = "1", features = ["full"] }
bytes = "1"
```

`src/main.rs`:
```rust
use std::time::{Duration, Instant};
use bytes::Bytes;
use zeromq::{PullSocket, PushSocket, Socket, SocketRecv, SocketSend, ZmqMessage};

const ENDPOINT: &str = "tcp://127.0.0.1:18888";

#[tokio::main]
async fn main() {
    match std::env::args().nth(1).as_deref() {
        Some("push") => {
            let mut s = PushSocket::new();
            s.bind(ENDPOINT).await.unwrap();
            let p = Bytes::from(vec![b'x'; 128]);
            loop {
                if s.send(ZmqMessage::from(p.clone())).await.is_err() {
                    tokio::task::yield_now().await;
                }
            }
        }
        Some("pull") => {
            let mut s = PullSocket::new();
            s.connect(ENDPOINT).await.unwrap();
            tokio::time::sleep(Duration::from_millis(500)).await;

            let t0 = Instant::now();
            let deadline = t0 + Duration::from_secs(3);
            let mut count: u64 = 0;
            loop {
                let remaining = deadline.saturating_duration_since(Instant::now());
                if remaining.is_zero() { break; }
                match tokio::time::timeout(remaining, s.recv()).await {
                    Ok(Ok(_)) => count += 1,
                    _ => break,
                }
            }
            eprintln!("elapsed={:.3}s count={count}", t0.elapsed().as_secs_f64());
        }
        _ => {}
    }
}
```

Run:
```sh
cargo run --release -- push & sleep 0.3 && cargo run --release -- pull
```

Expected: ~1M messages in 3 s. Actual on 0.5/0.6: the process hangs after a
few thousand recvs. The loop never reaches the deadline, the inner `timeout`
never reports `Elapsed`, and the final `eprintln!` never fires.

Replacing the inner loop with a spawned recv task that runs to completion
(timing the window with a plain `tokio::time::sleep` outside it) restores
normal throughput on 0.6.

## Environment

- zeromq 0.6.0 (also reproduced on 0.5.0; not present on 0.4.1)
- tokio 1.x (full features), default multi-thread runtime
- Linux 6.12, x86_64

rgbkrk:
## Summary

- Bound each `FairQueue::poll_next` call to the ready events present at poll entry so wake events generated during polling are deferred to the executor's next poll.
- Deduplicate queued ready events by peer key to avoid repeatedly polling the same pending stream in one outer poll.
- Add focused unit coverage for the wake-then-pending path and integration coverage for `PullSocket::recv()` wrapped in `tokio::time::timeout`.

This cherry-picks Alexei Kornienko's fix from `alex/issue-248` with original authorship preserved.

Fixes #248.

## Testing

- `cargo test fair_queue -- --nocapture`
- `cargo test pull_recv_with_per_call_timeout_keeps_making_progress -- --nocapture`
- `cargo clippy --all --all-targets --no-default-features --features async-std-runtime,all-transport,async-dispatcher-macros -- --deny warnings`
- `git diff --check HEAD^ HEAD`

Note: local full `cargo fmt --check` currently reports an import-order diff in pre-existing `benches/codec.rs` from `origin/master`, outside this patch. GitHub's nightly formatting check passed on the first PR run.

### DIFF
--- a/src/fair_queue.rs
+++ b/src/fair_queue.rs
@@ -3,7 +3,7 @@ use futures::Stream;
 use parking_lot::Mutex;
 
 use std::cmp::Ordering;
-use std::collections::{BinaryHeap, HashMap};
+use std::collections::{BinaryHeap, HashMap, HashSet};
 use std::hash::Hash;
 use std::pin::Pin;
 use std::sync::atomic;
@@ -13,6 +13,7 @@ use std::task::{Context, Poll, Waker};
 pub(crate) struct QueueInner<S, K: Clone> {
     counter: atomic::AtomicUsize,
     ready_queue: BinaryHeap<ReadyEvent<K>>,
+    queued: HashSet<K>,
     streams: HashMap<K, Pin<Box<S>>>,
     waker: Option<Waker>,
     /// Callback invoked when a stream ends (peer disconnected).
@@ -23,10 +24,7 @@ pub(crate) struct QueueInner<S, K: Clone> {
 impl<S, K: Clone + Eq + Hash> QueueInner<S, K> {
     pub fn insert(&mut self, k: K, s: S) {
         self.streams.insert(k.clone(), Box::pin(s));
-        self.ready_queue.push(ReadyEvent {
-            priority: self.counter.fetch_add(1, atomic::Ordering::Relaxed),
-            key: k,
-        });
+        self.push_ready(k);
         if let Some(w) = &self.waker {
             w.wake_by_ref();
         }
@@ -34,6 +32,7 @@ impl<S, K: Clone + Eq + Hash> QueueInner<S, K> {
 
     pub fn remove(&mut self, k: &K) {
         self.streams.remove(k);
+        self.queued.remove(k);
     }
 
     /// Clear all streams and the ready queue.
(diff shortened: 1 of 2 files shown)


Alexei-Kornienko:
@rgbkrk can you confirm that it actually fixes the issue? Cause I was not able to test it locally on my current setup.

rgbkrk:
I'll open another PR to show failing on main.

Alexei-Kornienko:
@rgbkrk thanks. Merge it if you think it's ready

rgbkrk:
I'll admit, while it fixes it we may want to do some perf work to make sure that we're doing ok with the structures (but obviously hanging is worst performance).

PR #250, opened 2026-05-05
Add #248 timeout repro test

rgbkrk:
Adds only the #248 timeout repro test, just to try reproducing the failure before the fair-queue fix in #249.

### DIFF
--- a/tests/push_pull_timeout.rs
+++ b/tests/push_pull_timeout.rs
@@ -1,0 +1,172 @@
+#![cfg(feature = "tokio-runtime")]
+
+use bytes::Bytes;
+use zeromq::prelude::*;
+use zeromq::ZmqMessage;
+
+use std::net::TcpListener;
+use std::process::Stdio;
+use std::time::{Duration, Instant};
+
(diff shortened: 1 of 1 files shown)


tests/push_pull_timeout.rs
```diff
@@ -0,0 +1,172 @@
...
+    let count = received.expect("PullSocket::recv stopped making progress when wrapped in timeout");
+    println!("received {count} messages");
+    assert!(
+        count > 10_000,
+        "received too few messages before deadline: {count}"
+    );
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn pull_recv_with_per_call_timeout_keeps_making_progress_across_processes() {
```
rgbkrk, line 67: This test fails on `master`.

rgbkrk: It technically hangs and times out:

```
test pull_recv_with_per_call_timeout_keeps_making_progress_across_processes ... FAILED
test pull_recv_with_per_call_timeout_keeps_making_progress has been running for over 60 seconds
```

PR #251, opened 2026-05-14
Fan out peer sends concurrently

rgbkrk:
## Summary

- add a shared fanout helper that sends to matched peer queues concurrently
- use it for PUB, XPUB, and SUB/XSUB backend fanout/control broadcasts
- document the first OMQ integration slice and likely next chunks
- add a FairQueue cancel-safety regression test for pending recv futures

## Verification

- `cargo test test_fair_queue_recv_is_cancel_safe_after_pending_poll --lib`
- `cargo check --all-targets`
- `cargo test --test pub_sub`
- `cargo test --test xpub_sub`
- `cargo test --test xsub_xpub`

## CI Follow-up

- The first GitHub Formatting run failed on `benches/codec.rs` import order;
  this branch was amended to match the workflow's nightly rustfmt output.

### DIFF
--- a/docs/OMQ_INTEGRATION.md
+++ b/docs/OMQ_INTEGRATION.md
@@ -1,0 +1,36 @@
+# OMQ.rs Integration Notes
+
+Issue #240 tracks interest in pulling the useful parts of
+`github.com/paddor/omq.rs` into this crate without forcing a wholesale
+rewrite. OMQ is not a fork, so most pieces need to come across as narrow
+architectural changes rather than file-level copies.
+
+## Incorporated
+
+- Fanout sends now collect eligible peers and await their writes
+  concurrently. This follows the performance direction discussed in the issue:
+  a slow subscriber should not make PUB, XPUB, or XSUB application fanout wait
+  on every other peer sequentially.
+- The FairQueue recv primitive has a cancel-safety regression test. A pending
+  recv future can be dropped after polling without consuming the next message.
+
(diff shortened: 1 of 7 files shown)


Alexei-Kornienko (commented):
I don't like this variant of implementation cause it adds complexity (by introducing a new send queue structure) instead of removing it. 
I have 2 commits in my fork that improve this particular place:
1) https://github.com/Alexei-Kornienko/zmq.rs/commit/efe9be0aa7fc452cc339ad3ca1240c260bda8273
2) https://github.com/Alexei-Kornienko/zmq.rs/commit/ad1f631373b04955a41513bfa85492fba392bf3a

However I don't see a large performance gain even if we use join_all.

Alexei-Kornienko:
Goal of my change is not only to improve fanout case but also reduce total number of locks taken. So now we have 2 atomic operations for every message sent so 2*number of subscribers. In my variant we only have 2 atomic checks - 1 new connections and 1 for new subscriptions in the case when the state doesn't change. and then we don't need any locks for fanout messages

Alexei-Kornienko:
Also pls review my branch https://github.com/Alexei-Kornienko/zmq.rs/commits/alex/monoio-runtime/ if you have time. Basically I was working on the same thing (improving code structure to incorporate improvements from forks)

rgbkrk:
Awesome, can you open it as a PR? I'm going to close this one.

PR #252, opened 2026-05-14
Alex/monoio runtime

Alexei-Kornienko:
Goal of this MR is to compare peformance between runtimes and check the ease of adding another runtime.
Additionally it brings several optimizations in the existing socket logic.

Overall it seems quite large and I think If general approach is approved I'll split it into several smaller requests


Alexei-Kornienko:
One of the things I don't like in this request is changes in FairQueue waker. They are caused by the fact that monoio runtime requires different bound in terms of Send/Sync. I'm looking for a way to reduce the scope of changes related to fair queue

rgbkrk:
Thanks. Glad we're taking this all on.

PR #256, opened 2026-05-14
Borrow outbound messages during ZMTP encoding

quillaid:
Reduce internal outbound message cloning by encoding borrowed messages.

## What changed

- Changed internal ZMTP codec encoding to borrow `Message` values.
- Updated send, handshake, ROUTER, PUB/XPUB, and SUB/XSUB fanout call sites to
  pass borrowed outbound messages where possible.
- Kept public socket APIs unchanged.
- Added large multipart PUSH/PULL and DEALER/ROUTER send coverage.

This is clone/refcount reduction only. Payload bytes are still copied into the
contiguous encode buffer; gather-write/writev is separate follow-up work.

## Validation

- `cargo +nightly fmt --all -- --check`
- `cargo clippy --all --all-targets --no-default-features --features async-std-runtime,all-transport,async-dispatcher-macros -- --deny warnings`
- `cargo test --test send_multipart_large`
- `cargo test --lib`
- CI-style runtime `cargo test --all` variants during audit
- `cargo bench --no-run`

## Benchmark snapshot

Locked targeted Tokio/TCP runs:

- Baseline: `send-clone-before-targeted-20260514T215829Z`
- Candidate: `send-clone-after-targeted-20260514T215829Z`

The targeted slice showed mixed but promising wins, including a large
DEALER/ROUTER 4096B throughput improvement in that run. Full TCP+IPC standard
comparison against libzmq and pinned OMQ still needs to be repeated before
making broad performance claims.

## Status

Draft parked behind PR #252's codec-by-reference work. If #252 lands first,
this branch needs a manual rebase and may become partially redundant.

### DIFF
--- a/benches/codec.rs
+++ b/benches/codec.rs
@@ -25,6 +25,7 @@ fn bench_encode(c: &mut Criterion) {
     for &frames in MULTIPART_FRAME_COUNTS {
         for &size in FRAME_SIZES {
             let m = build_message(frames, size);
+            let message = Message::Message(m);
             let total_bytes = (frames * size) as u64;
             group.throughput(Throughput::Bytes(total_bytes));
             group.bench_with_input(
@@ -34,9 +35,7 @@ fn bench_encode(c: &mut Criterion) {
                     b.iter(|| {
                         let mut codec = ZmqCodec::new();
                         let mut dst = BytesMut::with_capacity(total_bytes as usize + 64);
-                        codec
-                            .encode(Message::Message(m.clone()), &mut dst)
-                            .expect("encode");
+                        codec.encode(&message, &mut dst).expect("encode");
                         black_box(dst);
                     });
                 },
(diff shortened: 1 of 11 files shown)


benches/codec.rs
```diff
@@ -25,6 +25,7 @@ fn bench_encode(c: &mut Criterion) {
     for &frames in MULTIPART_FRAME_COUNTS {
         for &size in FRAME_SIZES {
             let m = build_message(frames, size);
+            let message = Message::Message(m);
```
poyea, line 28: This seems to conflict with https://github.com/zeromq/zmq.rs/pull/252/changes#diff-dddc0c5dc69fd0f65fd4c10886dd209eeb9f749a83b03d958f1c4d2bea9d7fb2R29

quillaid: Agreed, the overlap with #252 is real. I removed the unrelated formatting churn and updated the PR body to mark this as a draft/parked branch. If #252 lands first, this needs a manual rebase and may become partially redundant; I am not asking for merge review on it yet.

quillaid:
Parking this draft for now while I reduce the number of open perf-farm PRs and re-audit the stack. The overlap with #252 makes this a poor review target today. The idea can come back later after rebasing against the current codec work and rerunning the full validation/benchmark loop.

quillaid:
Scope audit after parking:

This PR is not primarily a stacking problem. The issue is that the change is cross-cutting across the outbound send/encode paths and overlaps active codec work in #252. Before reopening, it should be rebased against the current codec changes, checked for redundancy, and backed by focused clone/allocation or throughput evidence. If the surviving diff stays to borrowed outbound encoding plus the large-send coverage, it can be a coherent PR.

PR #260, opened 2026-05-14
Plan inproc transport lifecycle path

quillaid:
Add planning and lifecycle-test groundwork for a future `inproc://` transport.
This does not add public inproc endpoint parsing or a byte-stream prototype.

## What changed

- Added `docs/INPROC_TRANSPORT_PLAN.md`, covering the current transport
  boundary, registry/lifecycle requirements, teardown behavior, and correctness
  traps for a real inproc path.
- Recorded relevant libzmq and pinned OMQ behavior, including registry scope,
  duplicate bind handling, connect-before-bind semantics, and OMQ's codec-less
  inproc fast path.
- Added bind-map lifecycle tests in `tests/connect.rs` for duplicate active TCP
  binds, duplicate active IPC binds, and `NoSuchBind` on unbound `unbind`.

## Why this stops short of a prototype

A byte-stream-only inproc transport would go through the existing `FramedIo`
and ZMTP codec/greeting path, preserving the overhead this work is meant to
avoid. The plan keeps the public API unchanged until the core/Sans-I/O boundary
can support a no-regret implementation.

## Validation

- `cargo test --test connect -- --test-threads=1`
- `git diff --check`
- GitHub Actions are green on this draft PR.

### DIFF
--- a/docs/INPROC_TRANSPORT_PLAN.md
+++ b/docs/INPROC_TRANSPORT_PLAN.md
@@ -1,0 +1,153 @@
+# Inproc Transport Plan
+
+This note maps the current transport boundary and records the constraints for
+an eventual `inproc://` path. It is intentionally a planning artifact: work
+package 7 in `PERFORMANCE_ROADMAP.md` says inproc should follow the core
+boundary cleanup, so this should not introduce a public endpoint or slow
+prototype before the lower layers are ready.
+
+## Current Transport Boundary
+
+The public socket API reaches transports through two functions:
+
+- `transport::begin_accept(endpoint, callback)` binds an endpoint, starts an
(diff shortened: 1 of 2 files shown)


quillaid:
Parking this draft for now while I reduce the number of open perf-farm PRs and re-audit the stack. The inproc lifecycle notes and tests are useful groundwork, but this is not an immediate performance patch. I will bring it back later only as a focused, validated inproc implementation path.

quillaid:
Scope audit after parking:

This PR is coherent as inproc groundwork: planning doc plus bind/unbind lifecycle tests. I parked it because it is not a concrete inproc transport implementation or an immediate performance win. If resumed, it should either be a small lifecycle-test PR or a focused inproc implementation slice, but not both plus broader design material.

PR #267, opened 2026-05-24
Improve socket send and PUB fanout performance

sunlei:
## Summary

This PR moves socket writes out of the foreground `send()` path and into bounded writer queues, then batches ready writes before flushing. The main goal is to reduce peer-backed send cost without changing the public socket API.

Main changes:

- add a shared batched writer queue helper for framed socket writes
- add per-peer writer queues and a single-peer sender fast path in the generic backend
- route ROUTER sends through the shared backend send helper
- move PUB fanout behind a local fanout queue with per-subscriber writer queues
- keep PUB publishing non-blocking for slow subscribers by dropping on full fanout or subscriber queues
- add codec, framed-read, and fair-queue hot-path improvements
- expand benchmarks so send admission, delivered latency, and batch throughput are measured separately

## Review guide

Suggested review order:

1. `src/write_queue.rs`: shared batching behavior and flush boundary.
2. `src/backend.rs` and `src/router.rs`: generic peer send path, single-peer cache, and ROUTER reuse of the backend helper.
3. `src/pub.rs`: PUB fanout queue, subscriber writer queues, and drop behavior for slow subscribers.
4. `src/codec/*`, `src/fair_queue.rs`, and `src/message.rs`: smaller hot-path changes.
5. `benches/*` and `docs/BENCHMARKS.md`: benchmark coverage and result interpretation.

The highest-risk semantic change is PUB send completion. The other socket send paths still preserve local queue backpressure.

## Behavior notes

`PubSocket::send().await` is intentionally lighter after this change. It now means the message was accepted into the local PUB fanout path, or dropped according to PUB policy. It is not a delivery or transport-flush barrier.

For non-PUB peer-backed sockets, `send().await` still preserves queue backpressure. The foreground path accepts the message into the local per-peer writer queue, and the background writer task performs framed I/O in queue order.

For each peer writer queue, accepted messages are still fed to the framed sink in queue order. Batching changes flush granularity, not per-peer ordering for messages that were accepted into the queue.

## Benchmark scope

The benchmark suite now separates three different questions:

- send-admission cost: how quickly `send()` accepts a message into the local socket path
- delivered latency: how long until a peer observes a single message
- batch throughput: sustained pipeline throughput under batched sends

These are local macOS arm64 measurements. They are meant to show direction and order of magnitude; network benchmarks are scheduler-sensitive.

### 256B before/after socket-send cost

[...]

PR #268, opened 2026-05-25
Add hotpath benchmark coverage

sunlei:
## Summary

This splits the benchmark-only part out of #267 so it can be reviewed independently.

Changes:

- add a `hotpath` benchmark target for send-admission and delivered-latency comparisons
- add environment-based transport filtering for `throughput` and `compare_libzmq`
- extend benchmark documentation with benchmark families, result interpretation, and useful smoke commands
- keep the documentation limited to behavior that is true on current `master`; queued-send semantics will be documented in the later production PRs

## Validation

- `cargo fmt --check`
- `cargo bench --no-run`

## Follow-up

This PR is intentionally benchmark/documentation only. Production changes for queued sends, batching, PUB fanout, framed read, and FairQueue are split into separate PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,3 +70,7 @@ harness = false
 [[bench]]
 name = "throughput"
 harness = false
+
+[[bench]]
+name = "hotpath"
+harness = false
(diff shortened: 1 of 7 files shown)


Alexei-Kornienko (approved):
LGTM

PR #269, opened 2026-05-25
Optimize single-frame codec paths

sunlei:
## Summary

This splits the codec/message micro-optimizations out of #267 for focused review.

Changes:

- return single-frame decoded messages without first storing them in `buffered_message`
- write short and long frame headers into the destination buffer in one step
- keep public `ZmqMessage::from(Bytes)` on a lightweight single-frame construction path
- add unit coverage for the single-frame decode fast path and frame header encoding

## Validation

- `cargo fmt --check`
- `cargo test --lib codec::zmq_codec`

## Follow-up

This PR does not change socket send or receive scheduling. It is independent of the queued-send and PUB fanout PRs.

### DIFF
--- a/src/codec/zmq_codec.rs
+++ b/src/codec/zmq_codec.rs
@@ -5,7 +5,7 @@ use super::Message;
 use crate::ZmqMessage;
 
 use asynchronous_codec::{Decoder, Encoder};
-use bytes::{Buf, BufMut, Bytes, BytesMut};
+use bytes::{Buf, Bytes, BytesMut};
 
 use std::convert::TryFrom;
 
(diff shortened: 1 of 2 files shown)


Alexei-Kornienko:
@sunlei Don't think that this change actually worth merging. This might only introduce very minor gains or even no gains at all. Much better approach was used in one of the forks - They used a SmallVec to store message parts. This may allow us to reduce total number of heap allocations for most common messages (PUB socket messages, REQ/REQ/DEALER/ROUTER) they usually have a fixed number of parts that can be stored on the stack.

sunlei:
@Alexei-Kornienko Thanks, that makes sense.

I agree this PR is only a small codec-level cleanup, not a major performance fix. The local microbench numbers I had were:

- `codec/encode/frames=1/256`: about 7.6% faster
- `codec/encode/frames=2/4096`: about 6.8% faster
- `codec/decode/frames=1/4096`: `228.94 ns -> 218.86 ns`, about 2.9% faster

But I also confirmed that this does not materially change the end-to-end socket benchmarks. So I agree this is not the main bottleneck.

The `SmallVec` direction is a better larger fix. I had avoided it in this split because `ZmqMessage` currently exposes `VecDeque`-shaped APIs (`iter()` returns `VecDeque::Iter`, and there is `into_vecdeque()` / `TryFrom<VecDeque<Bytes>>`), so changing the internal representation looks like a separate API/design PR rather than a small codec PR.

I'm happy to close this PR and follow up separately with a focused message representation proposal using `SmallVec` or a similar small-message representation, if that is the preferred direction.

Feel free to close this PR at any time.

Alexei-Kornienko:
Agreed to close this in favor of future PR that would refactor ZmqMessage to use SmallVec to store Bytes

PR #270, opened 2026-05-25
Increase framed read chunk size

sunlei:
## Summary

This splits the framed read change out of #267 so it can be reviewed separately from send batching and FairQueue changes.

Changes:

- replace the default `asynchronous-codec` framed reader with `ZmqFramedRead`
- use a 64 KiB read scratch buffer instead of the default 8 KiB transport read chunk
- keep the existing `ZmqCodec` decode and EOF handling behavior
- add regression coverage for buffered consecutive messages and truncated frames at EOF

## Validation

- `cargo fmt --check`
- `cargo test --lib test_zmq_framed_read`

## Follow-up

This PR is receive-path only. It does not include FairQueue scheduling changes or writer queue batching.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,10 @@ name = "codec"
 harness = false
 
 [[bench]]
+name = "framed_read"
+harness = false
+
+[[bench]]
 name = "compare_libzmq"
 harness = false
 
--- a/benches/framed_read.rs
+++ b/benches/framed_read.rs
@@ -1,0 +1,116 @@
+//! Framed-read microbenchmarks for receive-path chunk sizing.
+
+#[allow(dead_code)]
+mod bench_runtime;
+
+use asynchronous_codec::{Encoder, FramedRead};
+use bytes::{Bytes, BytesMut};
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
+use futures::{executor::block_on, io::Cursor, Stream, StreamExt};
+
+use zeromq::{
+    __bench::{zmq_framed_read, Message, ZmqCodec, ZmqFramedRead},
+    ZmqMessage,
+};
+
+const FRAME_SIZES: &[usize] = &[256, 4096, 65536];
+const MESSAGE_COUNTS: &[usize] = &[1, 1024];
+
+const GREETING_STUB: [u8; 64] = {
+    let mut greeting = [0; 64];
+    greeting[0] = 0xff;
+    greeting[9] = 0x7f;
+    greeting[10] = 3;
+    greeting[11] = 1;
+    greeting[12] = b'N';
+    greeting[13] = b'U';
+    greeting[14] = b'L';
+    greeting[15] = b'L';
+    greeting
+};
+
+fn encoded_stream(message_count: usize, frame_size: usize) -> Vec<u8> {
+    let mut stream = BytesMut::new();
+    stream.extend_from_slice(&GREETING_STUB);
+
+    let payload = Bytes::from(vec![0xAB; frame_size]);
+    let message = Message::Message(ZmqMessage::from(payload));
+    let mut codec = ZmqCodec::new();
+
+    for _ in 0..message_count {
+        codec
+            .encode(message.clone(), &mut stream)
+            .expect("encode framed-read input");
+    }
+
+    stream.to_vec()
+}
+
+async fn drain_stream<S, E>(mut stream: S) -> usize
+where
+    S: Stream<Item = Result<Message, E>> + Unpin,
+    E: std::fmt::Debug,
+{
+    let mut messages = 0;
+
+    while let Some(item) = stream.next().await {
+        if matches!(item.expect("decode framed-read input"), Message::Message(_)) {
+            messages += 1;
+        }
+    }
+
+    messages
+}
+
+fn bench_framed_read(c: &mut Criterion) {
+    let mut group = c.benchmark_group("framed_read/drain");
+    bench_runtime::configure_group(&mut group);
+
+    for &message_count in MESSAGE_COUNTS {
+        for &frame_size in FRAME_SIZES {
+            group.throughput(criterion::Throughput::Bytes(
+                (message_count * frame_size) as u64,
+            ));
+
+            let input = encoded_stream(message_count, frame_size);
+            group.bench_with_input(
+                BenchmarkId::new(format!("default/messages={message_count}"), frame_size),
+                &input,
+                |b, input| {
+                    b.iter_batched(
+                        || Cursor::new(input.clone()),
+                        |cursor| {
+                            let reader = FramedRead::new(cursor, ZmqCodec::new());
+                            let messages = block_on(drain_stream(reader));
+                            assert_eq!(messages, message_count);
+                            black_box(messages);
+                        },
+                        BatchSize::SmallInput,
+                    );
+                },
+            );
+
+            group.bench_with_input(
+                BenchmarkId::new(format!("adaptive/messages={message_count}"), frame_size),
+                &input,
+                |b, input| {
+                    b.iter_batched(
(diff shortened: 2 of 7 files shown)


Alexei-Kornienko:
Don't like this PR for following reasons:
1) It creates a large read buffer by default. While it might be suitable for your particular scenario it might be an overkill for some tiny process the doesn't need to handle large number of messages. So a buffer size should be configurable and small by default
2) you introduce a separate scratch buffer to read data into and later copy parts of this buffer into BytesMut. This introduces additional copy of data.

So overall I don't see how this PR will improve performance or usability of the library.
My overall idea in terms of codec optimization would be to remove data copy as much as possible and reduce overall memory footprint in case of low throughput.
In terms of FramedRead improvements I would probably use the following approach:
1) Use a small BytesMut buffer by default and read directly into it (removes unnecessary data copy) return slices of this buffer as Bytes in final messages (basically that's how it works now)
2) Overtime this buffer might grow if we have a lot of messages in flight and memory is not reclaimed
3) When this buffer grows naturally over a certain threshold (64KB for example) allocate a new buffer with 64Kb capacity and start reading new messages into it (the old buffer should be freed automatically when all in-flight messages would be dropped) This will allow us to reduce the total number of heap allocation calls and possibly reduce memory fragmentation (need to check how it would actually perform)

sunlei:
@Alexei-Kornienko 

Thanks, I agree with the design concerns.

The motivation for this PR was a large-payload receive-path benchmark, not a general memory-footprint improvement. I was trying to isolate the remaining IPC 4096B receive-path gap after the send-side batching changes. The benchmark shape was roughly:

- one-way `DEALER/ROUTER`: DEALER sends `BATCH_SIZE` messages, ROUTER receives `BATCH_SIZE` messages, no echo/reply
- one-way `PUSH/PULL`: PUSH sends `BATCH_SIZE` messages, PULL receives `BATCH_SIZE` messages
- payload size: mostly `4096B`, with some `256B` checks
- goal: isolate common receive/framed-read/FairQueue cost from ROUTER identity and echo-send cost

The observation behind this PR was that `asynchronous-codec::FramedRead` reads from the underlying transport in relatively small chunks. For a `1024 * 4096B = 4 MiB` one-way receive batch, an 8 KiB read chunk means many more read/poll iterations than needed. So I tried a local `ZmqFramedRead` with a 64 KiB read chunk.

Local benchmark data from that experiment:

| workload | before | after 64 KiB read chunk | result |
|---|---:|---:|---|
| `zmqrs D/R one-way ipc 4096` | `4.1962 ms` | `3.8797 ms` | about 9.1% faster |
| `zmqrs P/P one-way ipc 4096` | `4.1945 ms` | `3.9731 ms` | about 5.8% faster |
| `zmqrs D/R one-way ipc 256` | around `525 us` before this pass | `478.88 us` | about 8.9% faster in that run |
| `zmqrs P/P one-way ipc 256` | around `486 us` before this pass | `495.78 us` | no meaningful improvement |

I also tried a 128 KiB variant. It did not improve D/R further, and P/P changes were within noise, so I dropped that and kept the more conservative 64 KiB version in my local performance branch.

For TCP large-payload smoke checks, the same branch still looked healthy:

| workload | zmqrs | libzmq |
|---|---:|---:|
| `D/R echo tcp 4096` | `1.9336 ms` | `11.094 ms` |
| `D/R one-way tcp 4096` | `1.0204 ms` | `4.2947 ms` |
| `P/P one-way tcp 4096` | `1.0332 ms` | `4.3525 ms` |

So there is a real benchmark reason why I tried this change, especially for large receive batches.

That said, I agree this implementation is too blunt for a library default:

1. It allocates a large buffer per reader, which is not appropriate for low-throughput or tiny-process use cases.
2. The separate scratch buffer means data is read into scratch first and then copied into `BytesMut`, which goes against the better direction of reducing copies.
3. It optimizes one large-payload profile without offering a configurable or adaptive memory policy.

Your suggested approach sounds like a better general-purpose design:

- start with a small `BytesMut` by default
- read directly into that `BytesMut` instead of going through a scratch buffer
- let the buffer grow naturally under load
- when it grows past a threshold, for example 64 KiB, rotate to a new buffer so old in-flight `Bytes` can release the old allocation naturally
- benchmark both low-throughput memory footprint and high-throughput large-payload receive performance

Given the tradeoff, I’m fine closing this PR instead of trying to merge this fixed-size read-buffer version. If this receive-path optimization is still useful to pursue, I can follow up separately with the adaptive `FramedRead` direction you described, including benchmarks for both memory footprint and large-payload receive throughput.

Alexei-Kornienko:
For your particular scenario I guess it would be fine to have a custom FramedRead impl and use a progressive chunk size. So lets say start with 64 bytes for a first read and if it's already available double the size for each next read until the code finds the sweet spot for itself. So my idea for this particular feature would look something like:

chunk_size: usize = 64;
buffer.reserve(chunk_size)
let bytes_read = socket.read(&mut buffer);
if bytes_read >= chunk_size { chunk_size *= 2;}

sunlei:
@Alexei-Kornienko

I reworked the follow-up implementation on `sunlei:split-framed-read-chunk` in the direction discussed here.

Implementation changes:

- Keep a custom `ZmqFramedRead`, but start with a small `64B` read chunk instead of a large default allocation.
- Remove the separate scratch read buffer. The reader now reads directly into `BytesMut` and trims it back to the actual number of bytes read.
- Grow the chunk only after a full read, capped at `64 KiB`.
- Add a small decoder-demand hint: after the ZMTP frame length is known, the reader uses `max(current_chunk, bytes_needed_by_decoder).min(64 KiB)`. This keeps the progressive behavior, but avoids excessive `64 -> 128 -> 256 -> ...` reads for a single large frame.
- Add a focused `framed_read` Criterion benchmark comparing the default `asynchronous_codec::FramedRead` path against the adaptive `ZmqFramedRead` path.

Longer local benchmark command:

```console
ZMQRS_BENCH_SAMPLE_SIZE=20 \
ZMQRS_BENCH_WARMUP_MS=2000 \
ZMQRS_BENCH_MEASUREMENT_MS=10000 \
cargo bench --bench framed_read
```

Results below are the Criterion point estimates from the same local run. This benchmark uses an in-memory `Cursor`, so it isolates framed-read/decode behavior rather than full TCP/IPC socket behavior.

| workload | default `FramedRead` | adaptive `ZmqFramedRead` | time reduction |
|---|---:|---:|---:|
| `1 msg / 256B` | `367.48 ns` | `344.53 ns` | `6.2%` |
| `1 msg / 4096B` | `538.97 ns` | `443.33 ns` | `17.7%` |
| `1 msg / 65536B` | `3.4692 us` | `3.1527 us` | `9.1%` |
| `1024 msg / 256B` | `53.784 us` | `53.120 us` | `1.2%` |
| `1024 msg / 4096B` | `370.09 us` | `312.07 us` | `15.7%` |
| `1024 msg / 65536B` | `4.2963 ms` | `2.6206 ms` | `39.0%` |

I also checked the hot path with `cargo asm`:

```console
cargo asm --lib '<zeromq::codec::framed::ZmqFramedRead as futures_core::stream::Stream>::poll_next'
```

`poll_read_into_buffer` is inlined into `poll_next`. The hot path now computes the read size from the adaptive chunk and the decoder demand, then clamps it to `64 KiB`. The previous `saturating_mul`-style growth path is gone; growth is a simple double-and-clamp after a full read.

This still does not implement the broader buffer-rotation/memory-reclamation idea from the earlier comment. It is intentionally scoped to the final progressive read chunk direction: small default footprint, direct reads into `BytesMut`, and adaptive chunk growth based on observed reads plus decoded frame demand.

Alexei-Kornienko (approved):
LGTM

sunlei:
@Alexei-Kornienko Thanks!
After the other two PRs are handled, I will continue to submit the remaining ones. The optimizations that have a more significant impact on performance are included in those upcoming PRs.

PR #272, opened 2026-05-25
Use shared bounded write queue for PUB

sunlei:
## Summary

This PR extracts the shared bounded writer helper and wires one concrete user into it: PUB subscriber writer tasks.

Changes:

- add `write_message_queue`, which feeds up to `128` immediately ready messages into `ZmqFramedWrite` and flushes once per bounded batch
- replace each PUB subscriber's locked `ZmqFramedWrite` with a bounded `mpsc::Sender<Message>`
- spawn one writer task per PUB subscriber that consumes the queue through `write_message_queue`
- keep the current foreground subscription scan in `PubSocket::send()`
- drop messages for full subscriber queues without backpressuring PUB send
- return writer `feed()` / `flush()` errors from `write_message_queue` to the owning writer task
- disconnect subscribers from the writer task when transport writes fail
- treat closed local queues as PUB drops on the send path, leaving cleanup to the writer/receiver side
- update the reverse PUB/SUB compliance test so blocking libzmq receiver joins do not block the async runtime while queued writer tasks flush
- add focused unit tests for topic matching, full queues, closed queues, writer errors, and batched writes

This PR intentionally does not add the later local PUB fanout queue or subscriber-count fast path. Those remain follow-up PRs.

## Semantics and tradeoffs

`PubSocket::send()` now enqueues messages into matching subscriber writer queues instead of awaiting each subscriber transport write while holding a per-subscriber async mutex. This keeps the publisher path independent from transport flush latency and lets the writer task batch bursts.

The tradeoff is that `Ok(())` no longer means the bytes have already been flushed to each subscriber transport. If a subscriber queue is full, PUB drops that message for that subscriber and still returns `Ok(())`, matching PUB slow-subscriber/drop behavior. If a local subscriber queue is closed, the send path treats it as a local PUB drop and continues fanout; transport error detection and subscriber cleanup are owned by the writer/receiver tasks.

`write_message_queue` now returns the first `feed()` / `flush()` error instead of swallowing it. The PUB writer task handles that error and calls `peer_disconnected()` for the affected subscriber.

## Benchmark context

The bounded writer data below comes from the original combined performance branch. In that experiment, moving writer tasks from per-message `send().await` / flush behavior to bounded `feed()` batches was the main TCP batch-throughput improvement.

| workload | before batching | after queued/batched writer | speedup vs before | libzmq comparison | speedup vs libzmq |
| `DEALER/ROUTER tcp 256` | `9.2927 ms` / `26.903 MiB/s` | `987.77 us` / `253.10 MiB/s` | `9.41x` | `1.4158 ms` / `176.57 MiB/s` | `1.43x` |
| `DEALER/ROUTER tcp 4096` | not in the same early baseline run | `2.7516 ms` / `1.4196 GiB/s` | n/a | `12.891 ms` / `310.28 MiB/s` | `4.68x` |

The PUB data below is also from the combined performance branch. This PR implements the per-subscriber writer queue piece; the local fanout queue and subscriber-count fast path are intentionally left for follow-up PRs, so these numbers should be read as motivation for the split series rather than as isolated #272-only results.

| workload | native before | combined branch result | speedup vs before | libzmq same/near run | combined branch vs libzmq |
| `PUB send-only 256B, 1 sub` | `6,428.8 ns` | `181.95 ns` | `35.33x` | `112.30 ns` | `1.62x` slower |
| `PUB send-only 256B, 4 subs` | `26,034.2 ns` | `192.87 ns` | `134.98x` | `110.76 ns` | `1.74x` slower |
| `PUB delivered latency 256B, 1 sub` | n/a | `28.735 us` | n/a | `85.851 us` | `2.99x` faster |
| `PUB delivered latency 256B, 4 subs` | n/a | `50.790 us` | n/a | `93.061 us` | `1.83x` faster |
| `PUB TCP throughput 256B, 1 sub` | n/a | `691.90 us` / `361.32 MiB/s` | n/a | `794.34 us` / `314.73 MiB/s` | `1.15x` faster |
| `PUB TCP throughput 256B, 8 subs` | n/a | `4.6646 ms` / `428.76 MiB/s` | n/a | `7.4259 ms` / `269.33 MiB/s` | `1.59x` faster |
| `PUB TCP throughput 4096B, 64 subs` | n/a | `76.861 ms` / `3.2526 GiB/s` | n/a | `465.15 ms` / `550.36 MiB/s` | `6.05x` faster |

## Validation

- `cargo fmt --check`
- `cargo test --lib`
- `cargo test --test pub_sub_compliant_reverse`
- `cargo test --test reconnection_compliant`

### DIFF
--- a/src/pub.rs
+++ b/src/pub.rs
@@ -27,29 +27,20 @@ fn subscription_matches(subscriptions: &[Vec<u8>], first_frame: &Bytes) -> bool 
     })
 }
 
-async fn send_to_subscriber(
-    backend: &PubSocketBackend,
-    peer_id: PeerIdentity,
-    mut send_queue: PubSendQueue,
-    message: &ZmqMessage,
-) -> ZmqResult<()> {
+fn send_to_subscriber(mut send_queue: PubSendQueue, message: &ZmqMessage) {
     let message = Message::Message(message.clone());
-    let result = match send_queue.try_send(message) {
-        Ok(()) => Ok(()),
+    match send_queue.try_send(message) {
+        Ok(()) => {}
         Err(error) if error.is_full() => {
             // PUB drops messages for slow subscribers so one full queue does not backpressure all publishers.
             drop(error.into_inner());
-            Ok(())
         }
-        Err(error) => Err(error.into_send_error()),
-    };
-    if result.is_err() {
-        // Sending to a disconnected subscriber is equivalent to dropping the message and pruning the stale link.
-        // One stale subscriber must not interrupt fanout of the current message to other subscribers.
-        backend.peer_disconnected(&peer_id);
+        Err(error) => {
+            // The writer task owns transport error detection and subscriber cleanup.
+            // A closed local queue is treated as a dropped PUB message on this send path.
+            drop(error.into_inner());
+        }
     }
-
-    Ok(())
 }
 
 pub(crate) struct Subscriber {
@@ -140,7 +131,18 @@ impl MultiPeerBackend for PubSocketBackend {
                 },
             )
             .await;
-        async_rt::task::spawn(write_message_queue(queue_receiver, send_queue));
+        let writer_backend = self.clone();
+        let writer_peer_id = peer_id.clone();
+        async_rt::task::spawn(async move {
+            if let Err(error) = write_message_queue(queue_receiver, send_queue).await {
+                log::debug!(
+                    "Error sending message to subscriber {:?}: {:?}",
+                    writer_peer_id,
+                    error
+                );
+                writer_backend.peer_disconnected(&writer_peer_id);
+            }
+        });
         let backend = self;
         let peer_id = peer_id.clone();
         async_rt::task::spawn(async move {
@@ -197,17 +199,12 @@ impl SocketSend for PubSocket {
             Some(frame) => frame,
             None => return Ok(()), // Empty message, nothing to publish
         };
-        let mut targets = Vec::new();
         let mut iter = self.backend.subscribers.begin_async().await;
         while let Some(subscriber) = iter {
             if subscription_matches(&subscriber.subscriptions, first_frame) {
-                targets.push((subscriber.key().clone(), subscriber.send_queue.clone()));
+                send_to_subscriber(subscriber.send_queue.clone(), &message);
             }
             iter = subscriber.next_async().await;
-        }
-
-        for (peer_id, send_queue) in targets {
-            send_to_subscriber(self.backend.as_ref(), peer_id, send_queue, &message).await?;
         }
         Ok(())
     }
@@ -322,7 +319,7 @@ mod tests {
     }
 
     #[async_rt::test]
-    async fn test_stale_subscriber_does_not_interrupt_fanout_to_matching_peer() {
+    async fn test_closed_subscriber_queue_does_not_interrupt_fanout_to_matching_peer() {
         let mut socket = PubSocket::new();
         let stale_peer = PeerIdentity::new();
         let live_peer = PeerIdentity::new();
@@ -358,7 +355,7 @@ mod tests {
         };
         assert_eq!(message.get(0), Some(&Bytes::from_static(b"payload")));
 
-        assert!(socket.backend.subscribers.get_sync(&stale_peer).is_none());
+        assert!(socket.backend.subscribers.get_sync(&stale_peer).is_some());
         assert!(socket.backend.subscribers.get_sync(&live_peer).is_some());
     }
 
@@ -392,15 +389,8 @@ mod tests {
         }
         assert!(prefilled_count > 0);
 
-        let result = send_to_subscriber(
-            &backend,
-            peer_id.clone(),
-            queue_sender,
-            &ZmqMessage::from("dropped payload"),
-        )
-        .await;
+        send_to_subscriber(queue_sender, &ZmqMessage::from("dropped payload"));
 
-        assert!(result.is_ok());
         assert!(backend.subscribers.get_sync(&peer_id).is_some());
 
         let mut queued_messages = Vec::new();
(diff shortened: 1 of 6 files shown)


Alexei-Kornienko:
For this PR I would actually prefer at least one usage example. Cause right now from the queue itself it's not clear how it supposed to be used and what are the pros/cons.
Could you please add a single socket (PUB for example) that would actually leverage this mechanism?

sunlei:
@Alexei-Kornienko I updated this PR to include a concrete PUB usage as requested.

The PR now wires `write_message_queue` into PUB subscriber writer tasks:

- each subscriber stores a bounded `mpsc::Sender<Message>` instead of a locked framed writer
- `peer_connected` spawns a writer task that drains that queue through `write_message_queue`
- `PubSocket::send()` still scans matching subscriptions in the foreground, but now enqueues into each matching subscriber writer queue
- full subscriber queues drop the message without backpressuring PUB send, while closed queues prune the stale subscriber and continue fanout

I also added focused tests for matching subscribers, full queues, stale subscribers, and the batched writer helper, plus adjusted the affected compliance tests so blocking libzmq receives do not block queued writer tasks. The latest CI run is green.

src/pub.rs
```diff
@@ -4,23 +4,57 @@ use crate::error::ZmqResult;
...
+    let result = match send_queue.try_send(message) {
+        Ok(()) => Ok(()),
+        Err(error) if error.is_full() => {
+            // PUB drops messages for slow subscribers so one full queue does not backpressure all publishers.
+            drop(error.into_inner());
+            Ok(())
+        }
+        Err(error) => Err(error.into_send_error()),
+    };
+    if result.is_err() {
```
Alexei-Kornienko: I'm not sure that this part is correct. Here you treat Error from mpsc channel as IO error and disconnect a subscriber.
My understanding is that  backend.peer_disconnected(&peer_id); should be actually called by the part that is doing IO (cause it can detect BrokenPipe error) and it should stop sending task and close channel from it's side. And this code should only handle it correctly.

src/pub.rs
```diff
@@ -4,23 +4,57 @@ use crate::error::ZmqResult;
...
+type PubSendQueue = mpsc::Sender<Message>;
+
+fn subscription_matches(subscriptions: &[Vec<u8>], first_frame: &Bytes) -> bool {
+    subscriptions.iter().any(|sub_filter| {
+        sub_filter.len() <= first_frame.len()
+            && sub_filter.as_slice() == &first_frame[0..sub_filter.len()]
+    })
+}
+
+async fn send_to_subscriber(
```
Alexei-Kornienko: guess this method doesn't need to be async cause I don't see any await here.

src/pub.rs
```diff
@@ -164,40 +200,14 @@ impl SocketSend for PubSocket {
         let mut targets = Vec::new();
         let mut iter = self.backend.subscribers.begin_async().await;
         while let Some(subscriber) = iter {
-            if subscriber.subscriptions.iter().any(|sub_filter| {
-                sub_filter.len() <= first_frame.len()
-                    && sub_filter.as_slice() == &first_frame[0..sub_filter.len()]
-            }) {
+            if subscription_matches(&subscriber.subscriptions, first_frame) {
                 targets.push((subscriber.key().clone(), subscriber.send_queue.clone()));
```
Alexei-Kornienko: If send to subscriber is sync and should be fast (doesn't do any locking) I guess we can call it here directly without the need to build a separate list of targets

Alexei-Kornienko (changes requested):
Mostly LGTM. However I think error handling should be improved and made more explicit (currently write task doesn't do any kind of error checks and it should be clear how IO errors and handled)

src/pub.rs
sunlei: Fixed in 9e497fe. `send_to_subscriber` no longer treats local mpsc send errors as IO errors and no longer calls `peer_disconnected()`. The send path now only handles enqueue/drop semantics. `write_message_queue` returns `feed()` / `flush()` errors, and the PUB writer task logs the transport error and disconnects the affected peer from the IO side.

src/pub.rs
sunlei: Fixed in 9e497fe. `send_to_subscriber` is now a synchronous helper because it only performs `try_send()` and has no await points.

src/pub.rs
sunlei: Fixed in 9e497fe. `PubSocket::send()` now calls `send_to_subscriber` directly while iterating matching subscribers, so the intermediate target list is gone.

sunlei:
@Alexei-Kornienko I pushed 9e497fe and resolved the review threads.

This update addresses the requested error-handling cleanup:

- `write_message_queue` now returns `feed()` / `flush()` errors to the owning writer task.
- the PUB writer task now logs transport write failures and calls `peer_disconnected()` from the IO side.
- `send_to_subscriber` is synchronous and only handles enqueue/drop behavior; local closed/full queues are treated as PUB drops on the send path.
- `PubSocket::send()` now calls `send_to_subscriber` directly during subscriber iteration, so the intermediate target list is removed.

CI is green on the latest run.

Alexei-Kornienko (approved):
LGTM

Alexei-Kornienko:
@rgbkrk Hi. Could you please also take a look on this one? Overall it should improve throughput for massive fanout scenarios by batching messages.

src/write_queue.rs
```diff
@@ -0,0 +1,153 @@
+use crate::codec::{CodecResult, Message, ZmqFramedWrite};
+
+use futures::channel::mpsc;
+use futures::{SinkExt, StreamExt};
+
+const WRITE_BATCH_LIMIT: usize = 128;
```
poyea, line 6: Out of curiosity how do we know this is optimal?

poyea (approved):

src/write_queue.rs
sunlei: I tested various parameters using benchmarks, but I did not save the data during the process. I believe it would be more practical to expose these as configurable parameters for more general use cases.

rgbkrk:
Landed now! Happy to look later this week.

PR #273, opened 2026-05-25
Queue generic backend writes

sunlei:
## Summary

This PR wires the generic multi-peer backend into the shared bounded writer queue that was added in #272.

Changes:

- store a bounded `mpsc::Sender<Message>` for each generic backend peer instead of storing `ZmqFramedWrite` directly
- spawn one writer task per peer that drains the queue through `write_message_queue`
- return writer `feed()` / `flush()` errors to the writer task and disconnect the affected peer from the I/O side
- keep non-PUB backpressure semantics by awaiting capacity when the selected peer queue is full
- document queued-send completion semantics on `SocketSend`
- update ROUTER/DEALER compliance tests so blocking libzmq thread joins do not block queued writer tasks from flushing

This PR intentionally does not add the single-peer sender cache or ROUTER send helper. Those remain follow-up work after this backend queueing step lands.

## Semantics and tradeoffs

For generic backend sockets, `send().await` now completes after the message has entered the selected peer's local writer queue. It does not mean the bytes have already been flushed to the transport.

Unlike PUB, generic backend sends still preserve backpressure. If the selected peer queue is full, the send path awaits queue capacity instead of dropping the message. If the queue is closed, the peer is disconnected and the send returns an error.

Transport write failures are handled by the peer writer task. The foreground send path only handles local queue acceptance/backpressure.

## Benchmark context

The data below comes from the original combined performance branch. This PR provides the generic backend queued-writer part of that series; later PRs still split out the single-peer cache and ROUTER helper, so these numbers should be read as motivation for the series rather than as isolated results from only this PR.

| workload | before queued/batched writer | combined branch result | speedup vs before | libzmq comparison | combined branch vs libzmq |
| `DEALER/ROUTER tcp 256` | `9.2927 ms` / `26.903 MiB/s` | `987.77 us` / `253.10 MiB/s` | `9.41x` | `1.4158 ms` / `176.57 MiB/s` | `1.43x` faster |
| `DEALER/ROUTER tcp 4096` | not in the same early baseline run | `2.7516 ms` / `1.4196 GiB/s` | n/a | `12.891 ms` / `310.28 MiB/s` | `4.68x` faster |

## Validation

- `cargo fmt --check`
- `cargo test --lib`
- `cargo test --lib backend`
- `cargo test --test router_dealer_compliant`
- `cargo test --test dealer_split`
- `cargo test --test router_split`
- `cargo test --test req_router_dealer_rep`
- `cargo test --test push_pull_timeout`

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -1,6 +1,8 @@
-use crate::codec::{FramedIo, Message, ZmqFramedRead, ZmqFramedWrite};
+use crate::async_rt;
+use crate::codec::{FramedIo, Message, ZmqFramedRead};
 use crate::fair_queue::QueueInner;
 use crate::util::PeerIdentity;
+use crate::write_queue::write_message_queue;
 use crate::{
     MultiPeerBackend, SocketBackend, SocketEvent, SocketOptions, SocketType, ZmqError, ZmqResult,
 };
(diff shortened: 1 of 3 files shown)


sunlei:
@Alexei-Kornienko Just a heads-up, this PR is currently blocking the subsequent ones.

Alexei-Kornienko (approved):
LGTM

PR #274, opened 2026-05-25
Queue PUB fanout locally

sunlei:
## Summary

This PR adds the local PUB fanout queue on top of the per-subscriber writer queues merged in #272.

Changes:

- add a bounded local fanout queue for `PubSocket::send()`
- move foreground PUB sends from subscriber-map scanning to local fanout enqueue
- drain the fanout queue in a background task that scans subscriptions and enqueues into matching subscriber writer queues
- coalesce fanout worker wakeups with a `wake_pending` flag
- add an atomic subscriber count so the no-subscriber path returns without touching the fanout queue
- keep PUB drop semantics when the local fanout queue or a subscriber writer queue is full
- preserve #272 writer-side transport error handling; closed local subscriber queues are still treated as PUB drops on the send path
- add focused tests for fanout queue drops, wake coalescing, no-subscriber fast path, subscriber count replacement/shutdown, topic matching, and closed/full subscriber queues

This PR intentionally does not include the generic backend single-peer cache or ROUTER helper.

## Semantics and tradeoffs

`PubSocket::send()` now returns after the message has entered the local fanout path, or after PUB drops it according to normal drop policy. It is not a delivery, subscriber enqueue, or transport flush barrier.

If there are no subscribers, `send()` returns immediately without enqueueing. If the local fanout queue is full, PUB drops the whole published message and returns `Ok(())`. During fanout, each matching subscriber still uses the per-subscriber writer queue from #272; a full subscriber queue drops that subscriber's copy without backpressuring the publisher.

The fanout worker owns subscription scanning. Per-subscriber transport write errors remain owned by each subscriber writer task.

## Benchmark context

The data below comes from the original combined performance branch. #272 already merged the per-subscriber writer queue; this PR adds the remaining local fanout queue and subscriber-count fast path from that PUB series. These numbers should be read as motivation for the PUB split series rather than as isolated results from only this PR.

| workload | native before | combined branch result | speedup vs before | libzmq same/near run | combined branch vs libzmq |
| `PUB send-only 256B, 1 sub` | `6,428.8 ns` | `181.95 ns` | `35.33x` | `112.30 ns` | `1.62x` slower |
| `PUB send-only 256B, 4 subs` | `26,034.2 ns` | `192.87 ns` | `134.98x` | `110.76 ns` | `1.74x` slower |
| `PUB delivered latency 256B, 1 sub` | n/a | `28.735 us` | n/a | `85.851 us` | `2.99x` faster |
| `PUB delivered latency 256B, 4 subs` | n/a | `50.790 us` | n/a | `93.061 us` | `1.83x` faster |
| `PUB TCP throughput 256B, 1 sub` | n/a | `691.90 us` / `361.32 MiB/s` | n/a | `794.34 us` / `314.73 MiB/s` | `1.15x` faster |
| `PUB TCP throughput 256B, 8 subs` | n/a | `4.6646 ms` / `428.76 MiB/s` | n/a | `7.4259 ms` / `269.33 MiB/s` | `1.59x` faster |
| `PUB TCP throughput 4096B, 64 subs` | n/a | `76.861 ms` / `3.2526 GiB/s` | n/a | `465.15 ms` / `550.36 MiB/s` | `6.05x` faster |

## Validation

- `cargo fmt --check`
- `cargo test --lib`
- `cargo test --lib pub`
- `cargo test --test pub_sub`
- `cargo test --test pub_sub_compliant`
- `cargo test --test pub_sub_compliant_reverse`
- `cargo test --test reconnection_compliant`
- `cargo test --test xpub_sub`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,3 +78,7 @@ harness = false
 [[bench]]
 name = "hotpath"
 harness = false
+
+[[bench]]
+name = "fanout_queue"
+harness = false
--- a/benches/fanout_queue.rs
+++ b/benches/fanout_queue.rs
@@ -1,0 +1,355 @@
+//! Fanout queue latency calibration for the PUB local fanout path.
+
+mod bench_runtime;
+
+use bench_runtime::BenchRuntime;
+use bytes::Bytes;
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use crossbeam_queue::ArrayQueue;
+use futures::channel::mpsc;
+use futures::StreamExt;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Instant;
+
+use zeromq::{__async_rt::task, ZmqMessage, ZmqResult};
+
+const FANOUT_QUEUE_CAPACITY: usize = 100_000;
+const THROUGHPUT_BATCH_MESSAGES: usize = 8_192;
+const MSG_SIZES: &[usize] = &[64, 256, 1024];
+
+struct ArrayFanoutQueue {
+    queue: Arc<ArrayQueue<ZmqMessage>>,
+    wake_sender: mpsc::UnboundedSender<()>,
+    wake_pending: Arc<AtomicBool>,
+}
+
+impl ArrayFanoutQueue {
+    fn publish(&self, message: ZmqMessage) -> ZmqResult<()> {
+        if self.queue.push(message).is_err() {
+            return Ok(());
+        }
+
+        if !self.wake_pending.load(Ordering::Acquire)
+            && !self.wake_pending.swap(true, Ordering::AcqRel)
+        {
+            let _ = self.wake_sender.unbounded_send(());
+        }
+
+        Ok(())
+    }
+}
+
+fn payload(size: usize, byte: u8) -> Bytes {
+    Bytes::from(vec![byte; size])
+}
+
+fn build_array_queue(capacity: usize) -> (ArrayFanoutQueue, mpsc::UnboundedReceiver<()>) {
+    let (wake_sender, wake_receiver) = mpsc::unbounded();
+    let queue = ArrayFanoutQueue {
+        queue: Arc::new(ArrayQueue::new(capacity)),
+        wake_sender,
+        wake_pending: Arc::new(AtomicBool::new(false)),
+    };
+    (queue, wake_receiver)
+}
+
+fn spawn_array_worker(
+    queue: Arc<ArrayQueue<ZmqMessage>>,
+    wake_pending: Arc<AtomicBool>,
+    mut wake_receiver: mpsc::UnboundedReceiver<()>,
+    ack_sender: mpsc::UnboundedSender<()>,
+) {
+    task::spawn(async move {
+        while wake_receiver.next().await.is_some() {
+            loop {
+                while let Some(message) = queue.pop() {
+                    black_box(message);
+                    let _ = ack_sender.unbounded_send(());
+                }
+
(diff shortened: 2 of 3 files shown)


sunlei:
@Alexei-Kornienko There are still a few more PRs to go before all the work for #267 is complete. Please take a look when you have a moment. Thanks a lot!

Alexei-Kornienko (commented):
I understand the general idea of the request however I don't understand why crossbeam_queue::ArrayQueue is needed. You already used genereal futures mpsc channel. My IMHO that it can be used instead without the need of adding more wakers and complicated logic. Just wait for new messages and process them in a separate task. What's the point of complicating things with ArrayQueue + wakers?

sunlei:
@Alexei-Kornienko The primary concern is performance. I will add a set of benchmarks to verify whether it is necessary to keep them.

sunlei:
@Alexei-Kornienko I pushed commit `8418d17` with a focused benchmark for this queue choice.

The benchmark compares the current `ArrayQueue + wake_pending` path with a bounded `futures::mpsc` channel across the cases that matter for this local fanout path:

- first-message wake latency: enqueue one message and wait for the worker ack
- enqueue return latency: producer-side queue admission while the worker is already pending
- full queue drop latency: PUB-style drop behavior when the local queue is full
- sustained throughput: send and drain batches of 8192 messages through a separate worker task

Command used locally:

```sh
ZMQRS_BENCH_SAMPLE_SIZE=20 \
ZMQRS_BENCH_WARMUP_MS=1000 \
ZMQRS_BENCH_MEASUREMENT_MS=3000 \
cargo bench --bench fanout_queue -- hotpath/pub_fanout_queue
```

Summary of the local results. The ratio column is `mpsc time / ArrayQueue time`, so values above `1.0x` mean the `ArrayQueue` path is faster:

| Case | Size | ArrayQueue | bounded `futures::mpsc` | Ratio |
|---|---:|---:|---:|---:|
| first-message wake latency | 64 B | 6.5674 us | 6.4161 us | 0.98x |
| first-message wake latency | 256 B | 7.3185 us | 7.3574 us | 1.01x |
| first-message wake latency | 1024 B | 7.5300 us | 7.5613 us | 1.00x |
| enqueue return latency | 64 B | 31.5507 ns | 54.7660 ns | 1.74x |
| enqueue return latency | 256 B | 31.6219 ns | 54.8724 ns | 1.74x |
| enqueue return latency | 1024 B | 31.8009 ns | 56.3525 ns | 1.77x |
| full queue drop latency | 64 B | 28.0570 ns | 29.1603 ns | 1.04x |
| full queue drop latency | 256 B | 27.8135 ns | 29.2999 ns | 1.05x |
| full queue drop latency | 1024 B | 27.8453 ns | 29.0453 ns | 1.04x |
| sustained throughput | 64 B | 1.1662 ms, 0.419 GiB/s | 1.9135 ms, 0.255 GiB/s | 1.64x |
| sustained throughput | 256 B | 1.3127 ms, 1.488 GiB/s | 2.0061 ms, 0.974 GiB/s | 1.53x |
| sustained throughput | 1024 B | 1.2751 ms, 6.127 GiB/s | 2.0354 ms, 3.838 GiB/s | 1.60x |

The first-message wake cost is effectively comparable. The current `ArrayQueue` path is materially better on producer-side enqueue latency, which is important for `PubSocket::send`, and it also has better sustained batch drain throughput in this isolated benchmark.

Also, `crossbeam_queue` is already used elsewhere in the crate, including `SegQueue`, so this does not introduce a new dependency family.

Validation:

```sh
cargo fmt --check
cargo bench --bench fanout_queue -- --test
cargo clippy --bench fanout_queue -- -D warnings
```

Alexei-Kornienko (approved):
LGTM

PR #276, opened 2026-05-27
Improve measured hot paths

sunlei:
## Summary

This PR contains a set of measured hot-path optimizations backed by focused Criterion benchmarks.

The changes are intentionally scoped to improvements that showed benchmark gains without stable regressions:

- Batch ZMTP frame header writes in `encode_frame`.
- Increase the adaptive framed-read initial chunk from 64 bytes to 128 bytes.
- Add a first-byte prefilter for PUB subscribers with multiple subscriptions.
- Add scoped `#[inline]` annotations for small hot-path functions with measured wins.
- Add focused benchmarks and benchmark-only helpers to cover these paths.

This PR does not add or change benchmark/release profiles. All numbers below were measured with the existing default bench profile.

### Codec frame header batching

`encode_frame` now writes short and long frame headers in a single contiguous operation:

- short frame: `extend_from_slice(&[flags, len])`
- long frame: stack `[u8; 9]` header followed by `extend_from_slice`

Assembly inspection confirmed that the short-frame header path no longer emits two separate `BufMut::put_slice` calls. The generated code now stores the two header bytes directly. The long-frame header path similarly becomes a direct byte store plus an unaligned 8-byte store.

| Benchmark | Baseline mean | PR mean | Latency reduction | Speedup |
|---|---:|---:|---:|---:|
| `codec/encode/frames=1/256` | `73.352 ns` | `68.688 ns` | `4.7168%` | `1.07x` |
| `codec/encode/frames=2/4096` | `207.31 ns` | `193.83 ns` | `6.5282%` | `1.07x` |
| `codec/encode/frames=8/4096` | `772.85 ns` | `754.56 ns` | `3.2473%` | `1.02x` |

### Adaptive framed-read initial chunk

`ZmqFramedRead` now starts with a 128-byte read chunk instead of 64 bytes.

This improves the direct framed-read drain benchmark for single-message workloads while keeping delivered-latency guards stable. A larger 256-byte initial chunk was also tested, but rejected because it regressed the large batched workload.

| Benchmark | Baseline mean | PR mean | Latency reduction | Speedup |
|---|---:|---:|---:|---:|
| `framed_read/drain/adaptive/messages=1/256` | `345.23 ns` | `247.91 ns` | `28.205%` | `1.39x` |
| `framed_read/drain/adaptive/messages=1/4096` | `487.97 ns` | `389.46 ns` | `21.013%` | `1.25x` |
| `framed_read/drain/adaptive/messages=1024/65536` | `2.6521 ms` | `2.6188 ms` | `1.2771%` | `1.01x` |

Delivered-latency guard results:

| Benchmark | Result |
|---|---|
| `hotpath/native_delivered_latency/push_pull/64` | no change |
| `hotpath/native_delivered_latency/push_pull/256` | no change |
| `hotpath/native_delivered_latency/push_pull/1024` | `3.0307%` noise-threshold improvement |
| `hotpath/native_delivered_latency/dealer_router/{64,256,1024}` | no change |

Rejected nearby candidate:

| Candidate | Reason rejected |
|---|---|
| `INITIAL_READ_CHUNK_SIZE = 256` | `framed_read/drain/adaptive/messages=1024/65536` regressed by `33.813%` |

### PUB multi-subscription first-byte prefilter

PUB subscription matching now keeps the single-subscription path unchanged, and only uses a first-byte prefilter when a subscriber has multiple subscriptions.

This avoids unnecessary full prefix comparisons for multi-subscription subscribers where most filters do not share the first byte with the message.

| Benchmark | Baseline mean | PR mean | Latency reduction | Speedup |
|---|---:|---:|---:|---:|
| `pub_fanout_backend/fanout_message_many_subscriptions/subs=1/filters=64/match_last` | `515.05 ns` | `341.71 ns` | `32.839%` | `1.51x` |
| `pub_fanout_backend/fanout_message_many_subscriptions/subs=1/filters=64/miss` | `384.48 ns` | `224.32 ns` | `42.992%` | `1.71x` |
| `pub_fanout_backend/fanout_message_many_subscriptions/subs=8/filters=8/match_last` | `1.4293 us` | `1.3007 us` | `8.4615%` | `1.10x` |
| `pub_fanout_backend/fanout_message_many_subscriptions/subs=8/filters=8/miss` | `630.77 ns` | `536.22 ns` | `15.967%` | `1.18x` |
| `pub_fanout_backend/fanout_message_many_subscriptions/subs=8/filters=64/match_last` | `2.9406 us` | `1.5683 us` | `46.435%` | `1.88x` |
| `pub_fanout_backend/fanout_message_many_subscriptions/subs=8/filters=64/miss` | `2.0740 us` | `748.31 ns` | `64.040%` | `2.77x` |

Original single-subscription and empty-subscription guard results:

| Benchmark | Result |
|---|---|
| `pub_fanout_backend/fanout_message_matches_all/subs=1/64` | no change |
| `pub_fanout_backend/fanout_message_matches_all/subs=8/64` | no change |
| `pub_fanout_backend/fanout_message_matches_all/subs=32/256` | no change |
| `pub_fanout_backend/fanout_message/subs=1/64` | `1.2828%` noise-threshold slowdown |
| `pub_fanout_backend/fanout_message/subs=8/256` | no change |
| `pub_fanout_backend/fanout_message/subs=32/256` | no change |

The single-subscription path is intentionally left in its original comparison shape. This optimization should be described as a many-subscription PUB fanout improvement, not as an empty-subscription or single-subscription improvement.

### Scoped hot-path inline annotations

This PR adds scoped `#[inline]` annotations only where focused benchmarks showed a stable improvement and guard benchmarks did not show a stable regression.

| Candidate | Benchmark evidence | Approx. speedup |
|---|---:|---:|
| `ZmqMessage::iter` | `message_accessors/iter_total_len/one_frame`: `64.801%` reduction | `2.84x` |
| `ZmqMessage::iter` | `message_accessors/iter_total_len/multipart`: `33.623%` reduction | `1.51x` |
| `From<Bytes> for ZmqMessage` | `message_construct/from_bytes/64`: `2.1279%` reduction | `1.02x` |
| `From<Bytes> for ZmqMessage` | `message_construct/from_bytes/256`: `2.3302%` reduction | `1.02x` |
| `From<Bytes> for ZmqMessage` | `message_construct/from_bytes/1024`: no change | `1.00x` |
| `TryFrom<VecDeque<Bytes>> for ZmqMessage` | `message_mutation_and_conversion/try_from_vecdeque_bytes`: `2.1168%` reduction | `1.02x` |
| `write_message_queue` | six direct writer batch workloads improved by `2.0146%` to `3.2632%` | `1.02x` to `1.03x` |
| `SubSocket::recv` | PUB delivered-latency cases improved by `3.9830%` to `11.404%` | `1.04x` to `1.13x` |
| `PullSocket::recv` | `native_delivered_latency/push_pull/1024`: `41.484%` reduction | `1.71x` |
| `PullSocket::recv` | `native_delivered_latency/push_pull/{64,256}`: no change | `1.00x` |

Notes:

- The `ZmqMessage::iter` and `From<Bytes>` wins are focused microbenchmark wins. They should not be interpreted as broad end-to-end socket latency wins.
- The `write_message_queue` evidence is from direct writer batch benchmarks.
- `PullSocket::recv` only showed a stable delivered-latency improvement for the 1024-byte push/pull case; 64-byte and 256-byte cases were no-change.

## Benchmark coverage added

This PR extends benchmark coverage for the changed paths:

| File | Coverage |
|---|---|
| `benches/hotpath.rs` | message construction/access/mutation, protocol conversions, FairQueue ready path, write queue, backend round-robin, PUB fanout, empty subscription, many subscriptions |
| `benches/compare_libzmq.rs` | XPUB/XSUB delivered-latency guard coverage for socket wrapper inline experiments |
| `src/lib.rs::__bench` | benchmark-only helpers for private/internal hot paths |

These helpers are exposed only under the existing benchmark-only internal module and are not intended as public API.

## Rejected candidates

The following nearby candidates were tested but not retained because they showed no stable benefit, unclear attribution, or stable regressions:

| Candidate | Reason rejected |
|---|---|
| codec single-frame decode fast path | single-frame improvement stayed below noise; multipart guard worsened |
| codec message pre-reserve | `codec/encode/frames=1/256` regressed |
| codec message pre-reserve with inline helper | inline did not fix the small-frame regression |
| `subscription_matches #[inline]` | empty-subscription guard regressed |
| `encode_frame #[inline]` | multipart large-message guard regressed |
| PUB single-subscriber move fast path variants | multi-subscriber guard regressed |
| backend single-peer cache removal | no stable benefit; `peers=16` slowed |
| codec decode iterative loop | `codec/decode/frames=1/256` regressed |
| PUB global `starts_with` rewrite | no stable improvement; 32-subscriber cases slowed |
| PUB empty-subscription local fast path | ordinary prefix guard regressed on repeat |
| codec encode single-frame branch | `codec/encode/frames=1/256` regressed by `1.5597%` |
| PUB multi-subscription `starts_with` inside prefilter | `filters=64` target workloads regressed |
| PUB fanout queue capacity `100_000 -> 131_072` | direct enqueue was no-change; sustained-throughput improvement had unclear attribution and changed buffering semantics |
| `[profile.bench]` changes | intentionally not included, to keep code-change benchmark evidence isolated |

## Validation

```text
cargo fmt --check
cargo check --all-targets
cargo bench --no-run
git diff --cached --check
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,11 +35,156 @@ pub mod __bench {
     //! DO NOT USE! PRIVATE IMPLEMENTATION, EXPOSED ONLY FOR BENCHMARKS.
     pub use super::codec::{CodecError, Message, ZmqCodec, ZmqFramedRead};
 
+    use super::backend::{GenericSocketBackend, Peer};
+    use super::codec::ZmqFramedWrite;
+    use super::r#pub::{PubSocketBackend, Subscriber};
+    use super::util::PeerIdentity;
+    use super::write_queue::write_message_queue;
+    use super::{SocketOptions, SocketType, ZmqMessage, ZmqResult};
+    use futures::channel::{mpsc, oneshot};
+    use futures::AsyncWrite;
+    use parking_lot::Mutex;
+    use std::sync::atomic::AtomicUsize;
+    use std::sync::Arc;
+
+    pub use super::fair_queue::FairQueue;
+
+    pub struct BenchRoundRobinBackend {
+        backend: GenericSocketBackend,
+        receivers: Vec<mpsc::Receiver<Message>>,
+    }
+
+    impl BenchRoundRobinBackend {
+        pub async fn new(peer_count: usize, queue_capacity: usize) -> Self {
+            let backend = GenericSocketBackend::with_options(
+                None,
+                SocketType::PUSH,
+                SocketOptions::default(),
+            );
+            let mut receivers = Vec::with_capacity(peer_count);
+
+            for peer_index in 0..peer_count {
+                let peer_id: PeerIdentity = format!("bench-peer-{peer_index}")
+                    .parse()
+                    .expect("valid peer identity");
+                let (send_queue, recv_queue) = mpsc::channel(queue_capacity);
+                backend
+                    .peers
+                    .upsert_async(peer_id.clone(), Peer { send_queue })
+                    .await;
+                backend.round_robin.push(peer_id);
+                receivers.push(recv_queue);
+            }
+
+            Self { backend, receivers }
+        }
+
+        pub async fn send_round_robin(&self, message: Message) -> ZmqResult<()> {
+            self.backend.send_round_robin(message).await
+        }
+
+        pub fn drain_ready(&mut self) -> usize {
+            self.receivers
+                .iter_mut()
+                .filter_map(|receiver| receiver.try_recv().ok())
+                .count()
+        }
+    }
+
+    pub struct BenchPubFanoutBackend {
+        backend: Arc<PubSocketBackend>,
+        receivers: Vec<mpsc::Receiver<Message>>,
+    }
+
+    impl BenchPubFanoutBackend {
+        pub async fn new(subscriber_count: usize, subscription: Vec<u8>) -> Self {
+            Self::new_with_subscriptions(subscriber_count, vec![subscription]).await
+        }
+
+        pub async fn new_with_subscriptions(
+            subscriber_count: usize,
+            subscriptions: Vec<Vec<u8>>,
+        ) -> Self {
+            let backend = Arc::new(PubSocketBackend {
+                subscribers: scc::HashMap::new(),
+                subscriber_count: AtomicUsize::new(0),
+                socket_monitor: Mutex::new(None),
+                socket_options: SocketOptions::default(),
+            });
+            let mut receivers = Vec::with_capacity(subscriber_count);
+
+            for subscriber_index in 0..subscriber_count {
+                let peer_id: PeerIdentity = format!("bench-sub-{subscriber_index}")
+                    .parse()
+                    .expect("valid peer identity");
+                let (send_queue, recv_queue) = mpsc::channel(1024);
+                let (stop_sender, _stop_receiver) = oneshot::channel();
+                backend
+                    .subscribers
+                    .upsert_async(
+                        peer_id,
+                        Subscriber {
+                            subscriptions: subscriptions.clone(),
+                            send_queue,
+                            _subscription_coro_stop: stop_sender,
+                        },
+                    )
+                    .await;
+                backend
+                    .subscriber_count
+                    .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+                receivers.push(recv_queue);
+            }
+
+            Self { backend, receivers }
+        }
+
+        pub async fn fanout_message(&self, message: ZmqMessage) {
+            self.backend.fanout_message(message).await;
+        }
+
+        pub fn drain_ready(&mut self) -> usize {
+            self.receivers
+                .iter_mut()
+                .filter_map(|receiver| receiver.try_recv().ok())
+                .count()
+        }
+    }
+
     pub fn zmq_framed_read<T>(inner: T) -> ZmqFramedRead
     where
         T: futures::AsyncRead + Unpin + Send + Sync + 'static,
     {
         ZmqFramedRead::new(Box::new(inner))
+    }
+
+    pub fn fair_queue_insert<S, K>(queue: &mut FairQueue<S, K>, key: K, stream: S)
+    where
+        K: Clone + Eq + std::hash::Hash,
+    {
+        queue.inner().lock().insert(key, stream);
+    }
+
+    pub fn fair_queue<S, K>(block_on_no_clients: bool) -> FairQueue<S, K>
+    where
+        K: Clone,
+    {
+        FairQueue::new(block_on_no_clients)
+    }
+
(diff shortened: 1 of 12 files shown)


src/lib.rs
Alexei-Kornienko, line 35: I would suggest to add a feature flag that would be disabled by default and that would control this module compilation. I don't like the fact that we ship code that is not actually used in production

Alexei-Kornienko (commented):
Minor question. Otherwise LGTM

Alexei-Kornienko:
@sunlei Thanks for the PR. If I understand correctly you are using zeromq in some of your projects and you try to improve it to fit your particular use case?
Can you describe a little bit more about what are your end goals and what you are trying to achieve? cause I have some prototype changes laying around for some time that can give a better improvement than trying to improve 2-3% of performance here and there.

src/lib.rs
sunlei: Addressed in 075099a. I added a disabled-by-default `bench-internals` feature, gated the `__bench` module with it, marked only the bench targets that depend on `__bench` with `required-features`, and updated the CI benchmark compile smoke to use `--features bench-internals`. This keeps benchmark-only internals out of normal production builds while preserving CI coverage for the benchmark targets.

sunlei:
@Alexei-Kornienko I am working on an HFT project where I use ZeroMQ for data transmission. Since latency is extremely critical and far more important than throughput, all my recent PRs have been focused on performance optimization.

My ultimate goal is to achieve the lowest possible latency. :)
I am very interested in learning about the new prototype you mentioned and am looking forward to seeing it.

Additionally, it would be even better if UDP support could be included.

Alexei-Kornienko (approved):
LGTM

Alexei-Kornienko:
@sunlei what socket pairs you are using and what type of async runtime?

Overall I pushed some WIP requests here - https://github.com/zeromq/zmq.rs/pull/252
And in several commits - https://github.com/zeromq/zmq.rs/pull/252/changes/ad1f631373b04955a41513bfa85492fba392bf3a
I try to reduce total number of locks taken/released for each message sent
And also here - https://github.com/zeromq/zmq.rs/pull/252/changes/202f1aae59aa99f7ac5a9c58e611cffb1ec8ff2f
I was experimenting with using writev for large messages to avoid extra copy of data. This gives a 30-40% performance boost for certain types of messages

Most of the work was done as an experiment with monoio runtime that uses io_uring for IO. But some part of the changes can be implemented in existing runtimes as well just need to spend some time on it.

sunlei:
@Alexei-Kornienko 

- I primarily use the `PUB/SUB` pattern.
- My runtime is based on `tokio`. I have considered using `io_uring`, but the legacy codebase is too extensive, making the migration cost prohibitively high.
- In my current scenario, the messages are quite small, with most being around 200B.

PR #265, opened 2026-05-15
Tune Tokio TCP socket buffers for PUSH/PULL throughput

quillaid:
## Summary

Tune TCP socket buffers for the Tokio transport path.

The default Tokio TCP transport now requests 4 MiB send and receive socket buffers when a connection is established or accepted. Buffer tuning is best-effort: connection setup continues if the platform refuses the requested size. Non-Tokio runtime builds keep the existing behavior.

## Performance

Focused locked Criterion evidence on the high-throughput PUSH/PULL TCP row, measured with the aggregate measurement stack from #261/#263:

- Baseline `a9c2d09`, `zmqrs/throughput/push_pull/tcp/peers=8/8192`: `1.43 GiB/s` estimate.
- Candidate measurement commit `d0db23e`, run `pushpull-tcp-buffers-d0db23e-20260515T0402Z`, same row: `3.80 GiB/s` estimate, interval `3.67-4.04 GiB/s`.
- Candidate sweep `pushpull-tcp-buffers-sweep-dirty-20260515T0354Z`, filter `zmqrs/throughput/push_pull/tcp`, completed all six rows without hangs. The target row repeated at `3.78 GiB/s`.

This PR isolates the transport change onto current master so the code diff stays limited to TCP buffer tuning.

OMQ remains ahead on the same comparison row at roughly `11-12 GB/s`, but this narrows the focused gap substantially without changing the public socket API or message semantics.

## Validation

- `cargo +nightly fmt --all -- --check`
- `cargo clippy --all --all-targets -- --deny warnings`
- `cargo clippy --all --all-targets --no-default-features --features async-std-runtime,all-transport,async-dispatcher-macros -- --deny warnings`
- `cargo test --all`
- `cargo test --all --no-default-features --features async-std-runtime,all-transport`
- `cargo test --all --no-default-features --features async-dispatcher-runtime,all-transport,async-dispatcher-macros`
- `cargo test --test connect --test large_message`
- locked `cargo bench --no-run`

Clippy emits existing renamed/removed configured-lint warnings but exits successfully.

## Risks

This is Tokio-only performance tuning. It increases requested per-connection TCP buffer capacity and relies on OS buffer limits, so exact performance impact may vary by platform and sysctl settings. The requests are best-effort to avoid changing connection success behavior. This should stay draft until CI and maintainer review confirm the new dependency/feature wiring is acceptable.

PR #279, opened 2026-05-29
Tune Tokio TCP socket buffers

rgbkrk:
## Summary

This creates a fresh, narrow PR inspired by the old #265 cloud-agent branch.

Changes:

- request 4 MiB TCP receive and send socket buffers for Tokio TCP connections
- apply the tuning on both outgoing `connect()` sockets and accepted listener sockets
- keep the tuning best-effort by logging failures at debug level and continuing connection setup
- leave non-Tokio runtime builds on the existing behavior through a no-op helper

## Notes

This intentionally preserves socket semantics and public API shape. The only behavior change is the best-effort OS socket-buffer request on the Tokio TCP transport path.

## Validation

```text
rustfmt --edition 2021 --check src/transport/tcp.rs
cargo check --features tokio-runtime,tcp-transport
cargo check --no-default-features --features async-std-runtime,tcp-transport
cargo check --no-default-features --features async-dispatcher-runtime,tcp-transport,async-dispatcher-macros
git diff --check
cargo test --test connect --test large_message
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ rust-version = "1.85.0"
 
 [features]
 default = ["tokio-runtime", "all-transport"]
-tokio-runtime = ["tokio", "tokio-util"]
+tokio-runtime = ["tokio", "tokio-util", "dep:socket2"]
 async-std-runtime = ["async-std"]
 async-dispatcher-runtime = ["async-std", "async-dispatcher"]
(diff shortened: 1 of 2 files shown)


rgbkrk:
Closing this duplicate in favor of #281. The two PRs have the same patch-id against current master, and #281 carries the fuller context/body for the TCP buffer tuning work.

PR #280, opened 2026-05-29
Trim dependency features and upgrade deps

sunlei:
## Summary

- Disable unnecessary production default features and keep dev dependencies broad for tests and benchmarks.
- Upgrade thiserror, rand, scc, and criterion.
- Adapt rand 0.10 and Criterion 0.7 API usage.
- Restrict endpoint port matching to ASCII digits after removing the regex unicode-perl feature.

## Validation

- cargo check
- cargo check --offline --no-default-features --features tokio-runtime,tcp-transport
- cargo check --no-default-features --features async-std-runtime,all-transport
- cargo check --no-default-features --features async-dispatcher-runtime,async-dispatcher-macros,all-transport
- cargo check --examples
- cargo test --no-run
- cargo test endpoint
- cargo bench --no-run
- cargo fmt --check

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,25 +25,29 @@ bench-internals = []
 
 [dependencies]
 async-dispatcher = { version = "0.1", optional = true }
-thiserror = "1"
-futures = "0.3"
+thiserror = "2"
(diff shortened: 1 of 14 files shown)


Alexei-Kornienko (approved):
LGTM

PR #285, opened 2026-05-29
Give libzmq comparison sockets buffer + io-thread parity

rgbkrk:
The libzmq baseline in the comparison benches ran with default socket buffers and a single I/O thread, while zmq.rs requests 4 MiB TCP send/recv buffers on every connection (#281) and runs its bench I/O on a 2-worker Tokio runtime. So the comparison handed zmq.rs two tuning advantages libzmq never got, and any "we narrowed the gap" reading, the #281 TCP-buffer result most of all, was partly an artifact of unequal tuning rather than a real win.

This gives the libzmq side parity:

- 4 MiB `SO_SNDBUF`/`SO_RCVBUF` on every libzmq socket, matching #281.
- `set_io_threads(2)` on every libzmq context, matching the bench's 2 Tokio workers.
- One generous high-water mark (100k) applied consistently, replacing the previously inconsistent per-callsite HWM (some sockets set none, others set 16k/100k).

## Design decision

The parity helpers live in the shared `bench_runtime` module (`libzmq_context`, `tune_libzmq_socket`) so both comparison benches configure libzmq the same way instead of drifting. HWM is set generous rather than matched, because zmq.rs exposes no public HWM knob; its internal queues are large, so a large libzmq HWM matches the intent (neither side throttled by a small in-flight window). The buffer size is a request that ipc ignores and only bites on tcp, which is exactly where zmq.rs tunes.

This levels the field; it does not flip results. libzmq still leads on the large throughput rows, leveling tuning widens that lead rather than reversing it. What it corrects is the set of "zmq.rs is competitive / closing in" claims that were measured against an under-tuned libzmq.

## Scope

Bench-only, no library code. This PR covers `compare_libzmq.rs` (the canonical comparison harness) and the shared `bench_runtime`. `throughput.rs` needs the same treatment but is being rewritten in #278, so that half lands there to avoid a conflict.

## Follow-up

A library gap surfaced during the audit: zmq.rs can't be tuned to match libzmq's HWM even when desired, because it has no public buffer/HWM API. Worth tracking separately.

### DIFF
--- a/benches/bench_runtime.rs
+++ b/benches/bench_runtime.rs
@@ -6,6 +6,45 @@ use criterion::{measurement::WallTime, BenchmarkGroup};
 #[allow(dead_code)]
 pub const DEFAULT_TRANSPORTS: &[&str] = &["tcp", "ipc"];
 
+// libzmq comparison parity. zmq.rs requests 4 MiB TCP send/recv buffers on
+// every connection (PR #281) and runs its bench I/O on a 2-worker Tokio
+// runtime, so the libzmq baseline has to be given the same buffers and io
+// threads or the comparison is skewed in zmq.rs's favor.
+#[allow(dead_code)]
+pub const LIBZMQ_TCP_BUFFER_BYTES: i32 = 4 * 1024 * 1024;
+/// Match the 2-worker bench Tokio runtime.
+#[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


Alexei-Kornienko (approved):
LGTM

PR #287, opened 2026-05-30
Guard ZMTP command parser length reads

### LINKED ISSUE #286: parsers in zmq.rs READY can panic
when zmq.rs handles READY, command length parser when the actual value is shorter than the length specified in the header. I think there should be simple check whether the actual length matches the specified length in the header. The same shape exists for the property name length . A single set of length checks before each split_to/buf[..n] will cover all three.

- found at: cbcb938a57d86bd8e8f1d94d8c3848e6d06321d1

- poc

  `Cargo.toml`:
  ```toml
  [package]
  name = "zmqrs-panic-poc"
  version = "0.1.0"
  edition = "2021"
  
  [dependencies]
  tokio = { version = "1", features = ["full"] }
  zeromq = { git = "https://github.com/zeromq/zmq.rs", rev = "cbcb938a57d86bd8e8f1d94d8c3848e6d06321d1" }
  ```
  
  `src/main.rs`:
  ```
  use std::error::Error;
  use tokio::io::{AsyncReadExt, AsyncWriteExt};
  use tokio::net::TcpStream;
  use tokio::time::{sleep, Duration};
  use zeromq::{RepSocket, Socket};
  
  fn greeting() -> [u8; 64] {
      let mut data = [0u8; 64];
      data[0] = 0xff;
      data[9] = 0x7f;
      data[10] = 3;
      data[11] = 0;
      data[12..16].copy_from_slice(b"NULL");
      data
  }
  
  #[tokio::main]
      let mut server = RepSocket::new();
      let endpoint = server.bind("tcp://127.0.0.1:0").await?;
      let addr = match endpoint.to_string().strip_prefix("tcp://") {
          Some(addr) => addr.to_string(),
          None => return Err("unexpected endpoint format".into()),
      };
  
      let mut peer = TcpStream::connect(addr).await?;
      let mut server_greeting = [0u8; 64];
      peer.read_exact(&mut server_greeting).await?;
      peer.write_all(&greeting()).await?;
  
      let mut server_ready = [0u8; 16];
      let _ = peer.read(&mut server_ready).await?;
  
      // Command frame, declared payload length 2. The first payload byte says
      // the command name is 5 bytes, so the parser indexes past the frame.
      peer.write_all(&[0x04, 0x02, 0x05, b'R']).await?;
      sleep(Duration::from_millis(300)).await;
  
      let _ = server.close().await;
      Ok(())
  }
  
  ```
  run with:
  ```
  cargo run
  ```

Alexei-Kornienko:
Fixes #286

### DIFF
--- a/src/codec/command.rs
+++ b/src/codec/command.rs
@@ -58,7 +58,15 @@ impl TryFrom<Bytes> for ZmqCommand {
     type Error = CodecError;
 
     fn try_from(mut buf: Bytes) -> Result<Self, Self::Error> {
+        if buf.is_empty() {
+            return Err(CodecError::Decode("Missing command length"));
+        }
+
         let command_len = buf.get_u8() as usize;
+        if buf.len() < command_len {
+            return Err(CodecError::Decode("Command name exceeds frame length"));
+        }
+
         // command-name-char = ALPHA according to https://rfc.zeromq.org/spec:23/ZMTP/
         let command = match &buf[..command_len] {
             b"READY" => ZmqCommandName::READY,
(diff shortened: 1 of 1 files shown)


Alexei-Kornienko:
@rgbkrk minor fix for the issue

rgbkrk:
Awesome, thank you.

PR #290, opened 2026-06-27
 Trim `syn 1` and `once_cell` from the build graph

Anthony-Eid:
I was looking at Zed's build graph and noticed this crate pulls in `syn 1` via async-std's `attributes` feature, even though it's mostly unneeded. That feature only provides the `#[async_std::main]` / `#[async_std::test]` macros re exported under `async-std-runtime`, but it was always enabled. Which caused all other runtimes to include   `async-attributes` and a `syn 1.x` as dependencies. This PR scopes `attributes` to the `async-std-runtime` feature, so that path keeps working while other runtimes don't include it.

I also noticed `once_cell` was a direct dependency used only for a couple of lazy regexes. Since the MSRV is already 1.85, I swapped those for the standard library's `std::sync::LazyLock` and dropped the dependency. 

These changes simplify the build graph and slightly improve build times.  Please let me know if you would like these changes to be in two separate PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,6 @@ scc = "3.7"
 crossbeam-queue = "0.3"
 uuid = { version = "1", default-features = false, features = ["v4"] }
 regex = { version = "1", default-features = false, features = ["std"] }
-once_cell = "1"
 log = "0.4"
 asynchronous-codec = "0.7"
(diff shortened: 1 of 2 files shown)


Alexei-Kornienko (approved):
LGTM

PR #292, opened 2026-07-19
Implement socket disconnect

NoMore201:
**Disclaimer**: AI was used to help with feature implementation (especially with writing unit tests and code review).

## Summary

Implemented `disconnect` and `disconnect_all` feature for all socket kinds.

- I tried to make API similar to the `binds`, `unbind`, `unbind_all` functions
- Added a new function `peer_list_by_endpoint` to `MultiPeerBackend` trait. This is the simplest solution I found to match an endpoint with its underlying peer ID.
- Added a new ZmqError enum `NoSuchConnection`

Second commit is a regression found by Claude consisting in a race condition between disconnect and reconnect features. If a disconnect is called while the reconnection tasks is performing a connection, we could end up with a connected but unreachable peer.

## Validation

```
cargo fmt --check
cargo check --all
cargo test --all
```

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -1,8 +1,8 @@
-use crate::async_rt;
 use crate::codec::{FramedIo, Message, ZmqFramedRead};
 use crate::fair_queue::QueueInner;
 use crate::util::PeerIdentity;
 use crate::write_queue::write_message_queue;
+use crate::{async_rt, Endpoint};
 use crate::{
     MultiPeerBackend, SocketBackend, SocketEvent, SocketOptions, SocketType, ZmqError, ZmqResult,
 };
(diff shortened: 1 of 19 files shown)


Alexei-Kornienko:
Do we really need this API? Is there some real life example when it would be useful?

NoMore201:
I don't have a real life example but I can describe my use case:

I have "consumer-producer" interfaces with an heartbeat mechanism. If one consumer interface loses HB, it forcefully disconnect and re-connects. This happens because interfaces have dynamic host/port assigned and negotiated with mDNS, so it could happen that an external node is rebooted, or simply the application that provides the interface restarted and it get assigned a new port. Without disconnet+reconnect I will never restore connection without manual intervention.

